PR #6, opened 2023-10-18
Add dev composer environment

RissyRan:
# Description

Add dev composer environment.
* Add dev env entry into Terraform.
* Add a helper function to check if entries are valid to BigQuery table.
* Add unit tests accordingly.

I think we still need to think about how we cooperate with various projects/datasets, i.e. should different team creates a separate Composer environment (gcs sync-up with the repo) or use the same one? For now, only prod and scheduled tests entries will be inserted to the table.

# Tests

* Unit tests
* Airflow run

* Unit tests for metric_test.py: [link](https://paste.googleplex.com/6281126409666560)
* Created Airflow: [link](https://screenshot.googleplex.com/446r4QeiqT6z4un)
* Airflow run in dev: [link](https://screenshot.googleplex.com/4cqs6PVvEbWSURT) 

# Checklist

### DIFF
--- a/deployment/cloud_composer.auto.tfvars
+++ b/deployment/cloud_composer.auto.tfvars
@@ -8,5 +8,9 @@ environment_config = [
   {
     environment_name   = "ml-automation-solutions"
     service_account_id = "ml-auto-solutions"
+  },
+  {
+    environment_name   = "ml-automation-solutions-dev"
+    service_account_id = "ml-auto-solutions-dev"
   }
 ]
--- a/apis/metric_config.py
+++ b/apis/metric_config.py
@@ -19,7 +19,7 @@ import enum
 from typing import Iterable, List, Optional
 
 
-# TODO(ranran): add project info to let uses specify dataset location
+# TODO(ranran): add project info to let users specify dataset location
 class DatasetOption(enum.Enum):
   BENCHMARK_DATASET = "benchmark_dataset"
   XLML_DATASET = "xlml_dataset"
--- a/configs/composer_env.py
+++ b/configs/composer_env.py
@@ -1,0 +1,45 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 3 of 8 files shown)


deployment/cloud_composer.auto.tfvars
```diff
@@ -8,5 +8,9 @@ environment_config = [
   {
     environment_name   = "ml-automation-solutions"
     service_account_id = "ml-auto-solutions"
+  },
+  {
+    environment_name   = "ml-automation-solutions-dev"
+    service_account_id = "ml-auto-solutions-dev"
```
will-cromar, line 14: Will the dev environment run all of the tests, or do we just expect people to upload one-offs?

RissyRan: Sorry that I missed this comment. Good question. I am expecting this env will be the one to test PR in CI/CD. Yes, I think we may still expect people to use their own composer env during the development. WDYT?

will-cromar: Do we have a way to stop automated tests from firing in the dev environment so we don't run two entire suites of tests?

People making their own composer environments for testing SGTM. It's hard to coordinate with one environment since the common code is all shared in the bucket.

RissyRan: Yes, I was thinking about this. One solution is to put a if/else statement in schedule flag

`schedule="0 6 * * *" if os.environ.get("COMPOSER_ENVIRONMENT") = "ml-auto-solutions" else None`

RissyRan: Actually, let me have a try and add into this PR.

RissyRan: Added `composer_env` file to hold constants and helper functions. 

I actually tested to update `SCHEDULED_TIME = "0 6 * * *" if composer_env.is_prod_env() else None` to a helper function, but Airflow scheduling does not recognize it. Also if I directly use `"0 6 * * *" if composer_env.is_prod_env() else None` inside of the dag, it throws too many arg error.

* Tested `metric_test.py` passed (plan to set it up in github): https://paste.googleplex.com/4535168667746304
* Uploaded to Airflow, and scheduled is None in dev: https://screenshot.googleplex.com/3JdYht3vcSUvmb6

will-cromar (approved):

PR #9, opened 2023-10-23
Move common attributes from TpuVmTest to TestConfig

RissyRan:
# Description

Move common attributes from TpuVmTest to TestConfig to address missing attributes issue in [JsonnetTpuVmTest](https://screenshot.googleplex.com/BqJCQ4tf7sgA7Kv)

# Tests

Upload the Airflow code, and no error is there. Then, trigger a test that is converted from Jsonnet.

Test run: [link](https://screenshot.googleplex.com/3wrcBh75atvQiha)

# Checklist

### DIFF
--- a/apis/task.py
+++ b/apis/task.py
@@ -48,6 +48,10 @@ class TpuTask(BaseTask):
     task_test_config: Test configs to run on this TPU.
     task_gcp_config: Runtime TPU creation parameters.
     task_metric_config: Metric configs to process metrics.
+    custom_tpu_name: A custom TPU name.
+    tpu_name_with_suffix: The flag to define if add auto-generated suffix.
+    all_workers: The flag to define if run commands on all workers or worker 0
+      only.
   """
 
   # TODO(wcromar): make these attributes less verbose
@@ -55,6 +59,9 @@ class TpuTask(BaseTask):
   task_gcp_config: gcp_config.GCPConfig
   tpu_create_timeout: datetime.timedelta = datetime.timedelta(minutes=60)
   task_metric_config: Optional[metric_config.MetricConfig] = None
+  custom_tpu_name: str = ''
+  tpu_name_with_suffix: bool = True
+  all_workers: bool = True
 
   def run(self) -> DAGNode:
     """Run a test job.
--- a/apis/task.py
+++ b/apis/task.py
@@ -97,14 +97,14 @@ class TpuTask(BaseTask):
     """
     with TaskGroup(group_id="provision") as group:
       with TaskGroup(group_id="initialize"):
-        if self.task_test_config.custom_tpu_name.strip():
-          base_tpu_name = self.task_test_config.custom_tpu_name
+        if self.custom_tpu_name.strip():
+          base_tpu_name = self.custom_tpu_name
         else:
           base_tpu_name = self.task_test_config.benchmark_id
 
         tpu_name = tpu.generate_tpu_name(
             base_tpu_name,
-            self.task_test_config.tpu_name_with_suffix,
+            self.tpu_name_with_suffix,
         )
         ssh_keys = ssh.generate_ssh_keys()
 
@@ -120,7 +120,7 @@ class TpuTask(BaseTask):
           # TODO(wcromar): remove split
           self.task_test_config.setup_script,
           ssh_keys,
-          self.task_test_config.all_workers,
+          self.all_workers,
       )
 
     return group, queued_resource_name, ssh_keys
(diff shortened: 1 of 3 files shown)


apis/test_config.py
```diff
@@ -100,11 +100,18 @@ class TestConfig(abc.ABC, Generic[A]):
...
+    custom_tpu_name: A custom TPU name.
+    tpu_name_with_suffix: The flag to define if add auto-generated suffix.
+    all_workers: The flag to define if run commands on all workers or worker 0
+      only.
   """
 
   accelerator: A
   time_out_in_min: Optional[int] = attrs.field(default=None, kw_only=True)
   task_owner: str = attrs.field(default='unowned', kw_only=True)
+  custom_tpu_name: str = attrs.field(default='', kw_only=True)
```
will-cromar, line 112: IMO defining TPU-specific attributes on `TestConfig` isn't the right solution. `TestConfig` should remain general, since it's going to be the parent of the GPU tests as well.

These make the most sense to me as attributes of `TpuTask`, since they're relevant to the case of dynamically creating a QR. They would not be relevant for GKE+TPU VM tests, since GKE controls execution and resource management.

will-cromar (commented):
We should be able to catch errors like this in the future by simple running all of the DAGs, or even running a type-checker like pyright or mypy on our code. The former sounds easier to set up, so I'll see what I can put together today.

apis/test_config.py
RissyRan: Yes, I think this makes more sense as well! Thanks for pointing it out! 

Tested: https://screenshot.googleplex.com/AvguLYU6aAEun6H

apis/task.py
```diff
@@ -48,13 +48,20 @@ class TpuTask(BaseTask):
     task_test_config: Test configs to run on this TPU.
     task_gcp_config: Runtime TPU creation parameters.
     task_metric_config: Metric configs to process metrics.
+    custom_tpu_name: A custom TPU name.
```
will-cromar, line 51: Can you add a note that the default is the task name?

apis/task.py
```diff
@@ -48,13 +48,20 @@ class TpuTask(BaseTask):
...
+    all_workers: The flag to define if run commands on all workers or worker 0
+      only.
   """
 
   # TODO(wcromar): make these attributes less verbose
   task_test_config: test_config.TestConfig[test_config.Tpu]
   task_gcp_config: gcp_config.GCPConfig
   tpu_create_timeout: datetime.timedelta = datetime.timedelta(minutes=60)
   task_metric_config: Optional[metric_config.MetricConfig] = None
+  custom_tpu_name: str = ''
```
will-cromar, line 62: nit: use `Optional[str]` with `None` as the default value instead

apis/task.py
```diff
@@ -90,14 +97,14 @@ def provision(self) -> Tuple[DAGNode, airflow.XComArg, airflow.XComArg]:
     """
     with TaskGroup(group_id="provision") as group:
       with TaskGroup(group_id="initialize"):
-        if self.task_test_config.custom_tpu_name.strip():
-          base_tpu_name = self.task_test_config.custom_tpu_name
+        if self.custom_tpu_name.strip():
```
will-cromar, line 100: Why `strip` here?

apis/task.py
```diff
@@ -48,13 +48,20 @@ class TpuTask(BaseTask):
     task_test_config: Test configs to run on this TPU.
     task_gcp_config: Runtime TPU creation parameters.
     task_metric_config: Metric configs to process metrics.
+    custom_tpu_name: A custom TPU name.
+    tpu_name_with_suffix: The flag to define if add auto-generated suffix.
```
will-cromar, line 52: WDYT of `suffix_tpu_name` instead?

will-cromar (approved):
Thanks!

Added a couple of minor nits about naming and annotations.

apis/task.py
RissyRan: Just in case anyone accidentally writes space for a custom name.

apis/task.py
RissyRan: Sounds good. Will update.

apis/task.py
RissyRan: Sure thing.

apis/task.py
RissyRan: Sounds good.

apis/task.py
RissyRan: shorter is better.

PR #5, opened 2023-10-16
Enable TensorFlow models on TPU multi-workers

RissyRan:
# Description

Enable TensorFlow models on TPU (with multi-workers):
* Enable to customize the TPU name, so that `--tpu=` config in Pod training can be known when passing run commands (I did not find a better way to obtain the tpu_name from XCom, as the value can only be retrieved within a task or python operator in Airflow).
* Write a few helper functions to construct model configs.
* Auto format the updated files.
* Other small changes, i.e. update args.

# Tests

Please describe the tests that you ran on Cloud VM to verify changes.

**Instruction and/or command lines to reproduce your tests:** 
Upload to an Airflow environment, and trigger a run.

**List links for your tests (use go/shortn-gen for any internal link):** 
* Successfully triggered the pod test and got the same error as single host (SSL issue: [b/304394983](https://buganizer.corp.google.com/issues/304394983))
* Verified logs as expected, i.e. `Connecting to IP addresses of worker 0...` in [pod training](https://screenshot.googleplex.com/4szQQHw36WQaR2N), `export TPU_LOAD_LIBRARY=0 `[env variable](https://screenshot.googleplex.com/AkFMaEkHqLndtPN) & [custom tpu name](https://screenshot.googleplex.com/8CQLWRzbwuyJhgN).

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run one-shot tests and provided workload links above if applicable. 
- [x] I have made or will make corresponding changes to the doc if needed.

PR #10, opened 2023-10-24
Add GitHub Action to load all DAG scripts locally

will-cromar:
# Description

Load DAGs to catch basic Python errors (syntax, missing attributes, etc).

Adds `requirements.txt` to use within GitHub Actions. This has to be manually kept in sync with Terraform unless we can read `requirements.txt` into our templates somehow.

Added some missing dependencies to `requirements.txt` and Terraform.

# Tests

Tested locally. This is expected to fail due to #5:

```
Traceback (most recent call last):
  File "/usr/local/google/home/wcromar/build/ml-auto-solutions/dags/xlml/pytorchxla_torchvision.py", line 45, in <module>
    ).run()
  File "/usr/local/google/home/wcromar/build/ml-auto-solutions/apis/task.py", line 69, in run
    provision, queued_resource, ssh_keys = self.provision()
  File "/usr/local/google/home/wcromar/build/ml-auto-solutions/apis/task.py", line 93, in provision
    if self.task_test_config.custom_tpu_name.strip():
AttributeError: 'JSonnetTpuVmTest' object has no attribute 'custom_tpu_name'
```

# Checklist

### DIFF
--- a/deployment/cloud_composer_template.tf
+++ b/deployment/cloud_composer_template.tf
@@ -105,10 +105,11 @@ resource "google_composer_environment" "example_environment" {
       # Note: keep this in sync with .github/requirements.txt
       pypi_packages = {
         apache-airflow-providers-sendgrid = ""
-        jsonlines                         = ""
         fabric                            = ""
-        google-cloud-tpu                  = ">=1.16.0"
+        google-cloud-bigquery             = ""
         google-cloud-storage              = ""
+        google-cloud-tpu                  = ">=1.16.0"
+        jsonlines                         = ""
         tensorflow-cpu                    = ""
       }
     }
--- a/deployment/cloud_composer_template.tf
+++ b/deployment/cloud_composer_template.tf
@@ -102,11 +102,14 @@ resource "google_composer_environment" "example_environment" {
       airflow_config_overrides = {
         core-allowed_deserialization_classes = ".*"
       }
+      # Note: keep this in sync with .github/requirements.txt
       pypi_packages = {
         apache-airflow-providers-sendgrid = ""
         jsonlines                         = ""
         fabric                            = ""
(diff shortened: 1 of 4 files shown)


will-cromar:
Moving to draft while I work out some GHA plumbing.

will-cromar:
Now the workflow fails where it's supposed to. Ready for review.

deployment/cloud_composer_template.tf
```diff
@@ -102,11 +102,15 @@ resource "google_composer_environment" "example_environment" {
       airflow_config_overrides = {
         core-allowed_deserialization_classes = ".*"
       }
+      # Note: keep this in sync with .github/requirements.txt
       pypi_packages = {
         apache-airflow-providers-sendgrid = ""
-        jsonlines                         = ""
         fabric                            = ""
+        google-cloud-bigquery             = ""
+        google-cloud-storage              = ""
```
RissyRan, line 110: I think we have google-cloud-bigquery & google-cloud-storage included?

https://cloud.google.com/composer/docs/concepts/versioning/composer-versions

RissyRan (commented):
Thanks Will! Just a minor comment.

deployment/cloud_composer_template.tf
will-cromar: Got it, that's why they weren't listed here. We do need to include them in `.github/requirements.txt`, so do you think they're worth including here too? Without a version specified, I expect this to be a no-op.

will-cromar: For some reason, Composer comes with the GPU version of TensorFlow by default, so `tensorflow-cpu` may cause issues. I'll comment these out and leave a note that these packages are pre-installed.

RissyRan (approved):
Thanks!

PR #15, opened 2023-11-01
Fix tpu name inconsistency in TF

RissyRan:
# Description

The [create_tpu_name](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/configs/xlml/tensorflow/solutionsTeam_tf_nightly_supported_config.py#L189) method is actually called 2 times: 1) initialize the tpu during [provision](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/configs/xlml/tensorflow/solutionsTeam_tf_nightly_supported_config.py#L171); and 2) trigger the training [process](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/configs/xlml/tensorflow/solutionsTeam_tf_nightly_supported_config.py#L146) in the run_model. 

* TPU name in provision: [link](https://screenshot.googleplex.com/8ebQYCv32GBZPZ2)
* TPU name in run_model: [link](https://screenshot.googleplex.com/BA5r6RxQxAFrqBS)

Set up variables in tpu name generation, and we could directly get it from helper methods (not need to be an operator or task).

# Tests

Run Airflow test and verify tpu names are the same

Single host:
* TPU name in `generate_tpu_name` task: `tf_resnet_imagenet-v2-8-69c6b471-e72b-4662-9fd0-bc7052cf5f30`
* TPU name in Airflow variable: key is `tf_resnet_imagenet-v2-8`, and value is `tf_resnet_imagenet-v2-8-69c6b471-e72b-4662-9fd0-bc7052cf5f30`.
* Able to successfully connect via SSH in [set_up and run_model]( https://screenshot.googleplex.com/87LjtZD78WURoTv) (failed due to other tf related issues, will be addressed in this [PR](https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/57))

Multi host:
* TPU name in `generate_tpu_name` task: `tf_resnet_imagenet-v4-32-4925fdbd-7f2d-4ff1-bcde-7ed1f2fecad6`
* TPU name in Airflow variable: key is `tf_resnet_imagenet-v4-32`, and value is `tf_resnet_imagenet-v4-32-4925fdbd-7f2d-4ff1-bcde-7ed1f2fecad6`.
* Able to successfully connect via SSH in [set_up and run_model]( https://screenshot.googleplex.com/4ExWYxfkg6a3G37)(failed due to other tf related issues)
* run_model retry has the same tpu name

# Checklist


configs/xlml/tensorflow/solutionsTeam_tf_nightly_supported_config.py
```diff
@@ -182,10 +185,12 @@ def export_env_variable(is_pod: bool) -> str:
...
+    return vm_resource.RuntimeVersion.TPU_VM_TF_NIGHTLY_POD.value
+  return vm_resource.RuntimeVersion.TPU_VM_TF_NIGHTLY.value
 
 
 def create_tpu_name(test_name: str, tpu_version: str, tpu_cores: int) -> str:
-  """Create a custom TPU name."""
-  return f"{test_name}-v{tpu_version}-{tpu_cores}-{str(uuid.uuid4())}"
+  """Create a custom TPU name with hash value of the current date."""
+  date_suffix = date.today().strftime("%Y%m%d")
+  date_hash = hashlib.sha256(str(date_suffix).encode("utf-8")).hexdigest()[:36]
```
will-cromar, line 195: Are you sure this works? I believe this part will run once when parsing the DAG, not each time the DAG is invoked.

will-cromar (commented):
Is the core issue here is that we need to build the run command including the TPU name statically (at DAG parsing time), but need to randomize the name dynamically (at run time)?

There should be a way to add dynamic data into the command if we don't have it yet. Does our SSH client give us a way to add environment variables to the target machine? If so, then it might make more sense to add an env var into the config to be expanded later.

configs/xlml/tensorflow/solutionsTeam_tf_nightly_supported_config.py
RissyRan: `I believe this part will run once when parsing the DAG` that was my understanding till Mark and I met TPU name consistency issue recently. Basically, there will be a different tpu_name when running the training process with `--tpu=tpu_name` (this is not affecting the SSH part, as we passed it as XCOM).

RissyRan:
> Is the core issue here is that we need to build the run command including the TPU name statically (at DAG parsing time), but need to randomize the name dynamically (at run time)?
> 
> There should be a way to add dynamic data into the command if we don't have it yet. Does our SSH client give us a way to add environment variables to the target machine? If so, then it might make more sense to add an env var into the config to be expanded later.

Yes, I think that's the issue. When writing helper function, we don't know what's the TPU name for the command like `python3 official/vision/train.py --tpu={tpu_name} --experiment=resnet_imagenet ... `. I tried to use xcom, but it mentioned missing task or operator during parsing. I haven't tried to set up environment variables, probably it's a way to go. Let me add a TODO to make the solution more elegant.

configs/xlml/tensorflow/solutionsTeam_tf_nightly_supported_config.py
RissyRan: I tried this code snippet

```
  tpu_name = create_tpu_name(test_name, tpu_version, tpu_cores)
  Variable.set(f"{test_name}_tpu_name", tpu_name)
  tpu_name_param = Variable.get(f"{test_name}_tpu_name") if is_pod else "local"

...

def create_tpu_name(test_name: str, tpu_version: str, tpu_cores: int) -> str:
  """Create a custom TPU name with hash value of the current date."""
#   date_suffix = date.today().strftime("%Y%m%d")
#   date_hash = hashlib.sha256(str(date_suffix).encode("utf-8")).hexdigest()[:36]
  return f"{test_name}-v{tpu_version}-{tpu_cores}-{str(uuid.uuid4())}"
```

Those are what I have observed:
* Variable is set up for ResNet v4-32: https://screenshot.googleplex.com/9oMW7xvCiMzKpYC
* TPU name is different: 
    * name from generate_tpu_name: tf_resnet_imagenet-v4-32-19b3586e-1b28-4686-91c5-ee99dedbf5b1
    * name from 1st run_model: tf_resnet_imagenet-v4-32-5d786247-dd9b-446b-8aaf-7e334c210ac4
    * name from 2nd run_model: tf_resnet_imagenet-v4-32-642ab819-4db8-4ff5-94b8-8f97cb4e8fcd

will-cromar: I would expect the name to change once every time the DAG parser runs, which runs every minute or two. I'm actually more surprised if anything that the name stays consistent for a  whole run, the DAG can actually get rebuilt while tasks are in progress. But if the command gets encoded as an XCOM somewhere, then that will stay consistent.

RissyRan: Yes, that's my current understanding. Like discussed offline, I will have a try with env variable in fabric run.

RissyRan: I changed back to the original implementation for `generate_tpu_name()`, but adding `Variable.set()` to set TPU name as an airflow variable. In this case, helper functions can easily get that TPU name value, and this is not needed to be in a task or operator like XCOM. Tested in the Airflow (including retry), and updated the description. 

I keep the current implementation in ssh, and think it's fine. So we could get TPU names in 2 ways: 1) for TF tensorflow tests, get the name through airflow viable by passing the benchmark ID as the key; or 2) use passed xcom value directly between tasks. Let me know your thoughts!

configs/xlml/tensorflow/solutionsteam_tf_nightly_supported_config.py
```diff
@@ -57,7 +59,10 @@ def get_tf_resnet_config(
   }
 
   test_name = "tf_resnet_imagenet"
-  tpu_name = create_tpu_name(test_name, tpu_version, tpu_cores)
+  benchmark_id = f"{test_name}-v{tpu_version}-{tpu_cores}"
+  # Add default_var to pass DAG check
+  tpu_name = Variable.get(benchmark_id, default_var=None) if is_pod else "local"
```
will-cromar, line 64: Are variables scoped to a given dag run?

From the [docs](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/variables.html) : `Variables are global, and should only be used for overall configuration that covers the entire installation; to pass data from one Task/Operator to another, you should use XComs instead.`

RissyRan: Yes, it's global variables, so I used benchmark_id. If I remember correctly, the problem for XCOMs is that it needs to be used in a task or an operator. So I haven't found a way to apply in `configs/xlml/tensorflow/solutionsteam_tf_nightly_supported_config.py` yet. Also mentioned in this comment: https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/15#discussion_r1433430158.

Let me know your thoughts!

will-cromar: I'm not sure this runs in the right order. The body of `generate_tpu_name` will run at the beginning of a DAG run, and this part will run during DAG construction.

If this does work, we can take it as a temporary hack to get the tests passing. But, please leave a TODO to replace this with an XCOM.

implementations/utils/tpu.py
```diff
@@ -33,13 +33,14 @@
...
 
 
 @task
-def generate_tpu_name(base_tpu_name: str, with_suffix: bool = True) -> str:
-  if with_suffix:
-    return f'{base_tpu_name}-{str(uuid.uuid4())}'
-  return base_tpu_name
+def generate_tpu_name(base_tpu_name: str) -> str:
+  tpu_name = f'{base_tpu_name}-{str(uuid.uuid4())}'
+  Variable.set(base_tpu_name, tpu_name)
```
will-cromar, line 42: Can you make setting the variable optional and non-default? That way this function remains pure in most cases. We also don't want references to global state like this to proliferate beyond this case.

will-cromar (approved):
To get this working through an XCOM, we would need some way to template runtime data into the command. Environment variables are probably the way to go, since the commands are already in a shell and everyone knows env vars.

This fix is straightforward, so let's just mark some TODOs to follow up.

configs/xlml/tensorflow/solutionsteam_tf_nightly_supported_config.py
RissyRan: Sure, I will leave a TODO and think about how to apply with XCOM in this helper config file. Thank you!

implementations/utils/tpu.py
RissyRan: Sounds good. Will update that.

PR #28, opened 2023-11-30
Enable docker image feature with xpk

RissyRan:
New [PR](https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/40).

# Description

Enable docker image feature with xpk:
* Add this as an GKE example, and plan to add more examples for quick start
* Add `TpuGkeTask` to execute run_model and post_process task groups 
* Add cluster configs under `configs/cluster` folder, so Airflow has the access to the created cluster with TPUs. It seems `certificate-authority-data` is safe to share based on this stackoverflow [thread](https://stackoverflow.com/questions/71444145/should-the-k8s-cluster-certificate-authority-be-kept-secret).
* Update `TpuTask` to `TpuGceTask` to distinguish `TpuGkeTask`, and similar for test config

A few TODO once b/311073979 is approved:
* Update GCP and cluster configs, and host clusters in cloud-ml-auto-solutions project
* Will have a separate PR to document how to provision cluster using XPK. It is quite straightforward with 1-2 commands

# Tests

* Upload codes to GCS bucket and test

* Airflow test: [link](https://screenshot.googleplex.com/4uVjRBGjNY5B6Q5)

# Checklist


dags/benchmark/pytorchxla_torchbench.py
```diff
@@ -31,9 +31,9 @@
 SCHEDULED_TIME = "0 17 * * *" if composer_env.is_prod_env() else None
 
 with models.DAG(
-    dag_id="pytorchxla-torchbench",
-    schedule=SCHEDULED_TIME,
-    tags=["pytorchxla", "nightly", "torchbench"],
+    dag_id="pytorch_nightly_torchbench",
+    schedule=None,
```
will-cromar, line 35: Is this intentional? Do we want to disable this test's schedule?

implementations/utils/xpk.py
```diff
@@ -0,0 +1,106 @@
...
+    accelerator_type: str,
+    run_cmds: str,
+    task_owner: str,
+    num_slices: int = 1,
+) -> KubernetesPodOperator:
+  """Run workload through xpk tool.
+
+  The reason to use KubernetesPodOperator instead of BashOperator is that
+  xpk must run with Python 3.10 or greater; however, the latest version in
+  Composer is Python 3.8, and it's non-trivial to upgrade it as the  Composer
```
will-cromar, line 48: :facepalm: I wish they would update this

Airflow actually provides a neat tool for this situation. You can run an `@task` directly in a Docker container with a different Python environment: https://airflow.apache.org/docs/apache-airflow/stable/tutorial/taskflow.html#dependency-separation-using-docker-operator

This gives us at least one less level of indirection: a docker container on the current host, rather than a docker container on another host. There's also a corresponding `@task.kubernetes` in case we do need a cluster!

implementations/utils/metric.py
```diff
@@ -397,8 +396,38 @@ def is_valid_entry() -> bool:
...
+  """Get job status for the GKE run.
+
+  FAILED - if any failure occurs in run_model
+  SUCCESS - end-to-end model tests are successful in run_model
+  """
+  context = get_current_context()
+  execution_date = context["dag_run"].logical_date
+  current_dag = context["dag"]
+
+  workload_completion = current_dag.get_task(
```
will-cromar, line 409: This DAG introspection looks kind of sketchy to me. Could the status here be plumbed through XCOM instead?

implementations/utils/xpk.py
```diff
@@ -0,0 +1,106 @@
...
+      owner=task_owner,
+  )
+
+
+@task.sensor(poke_interval=60, timeout=600, mode="reschedule")
+def wait_for_workload_completion(workload_id: str, cluster_config: str) -> bool:
+  """Check the workload status."""
+
+  # Load the config for the cluster with TPUs in the pool
+  config.load_kube_config(
```
will-cromar, line 84: If it's not too verbose, do you think it would be clearer to `import kubernetes` and write this as `kubernetes.config.etc`? Otherwise, the `config` and `client` modules are a little ambiguous

configs/example/gke_example_config.py
```diff
@@ -0,0 +1,61 @@
...
+    time_out_in_min: int,
+) -> task.TpuGkeTask:
+  # TODO(ranran): update the project once quota is approved (b/311073979).
+  job_gcp_config = gcp_config.GCPConfig(
+      project_name="tpu-prod-env-one-vm",
+      zone=tpu_zone,
+      dataset_name=metric_config.DatasetOption.XLML_DATASET,
+  )
+
+  run_model_cmds = (
```
will-cromar, line 37: If this test is actually going to be used, should it be in the `example` directory? If it is just an example, I would go even simpler and just print `jax.device_count()` to show the configuration is correct.

configs/cluster/v5e_cluster_config
```diff
@@ -0,0 +1,26 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
will-cromar: Okay, way back in ancient history we used to do something similar in the old implementation of the metrics handler. This is how we built the kube config programatically: https://github.com/GoogleCloudPlatform/ml-testing-accelerators/blob/3bd5720e667207fc6a49e8f692110fd812f4fc5b/metrics_handler/job_status_handler.py#L59-L94

apis/task.py
```diff
@@ -41,15 +41,15 @@ def run() -> DAGNode:
 
 
 @dataclasses.dataclass
-class TpuTask(BaseTask):
-  """This is a class to set up tasks for TPU.
+class TpuGceTask(BaseTask):
```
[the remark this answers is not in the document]
RissyRan: Yes, we could. If that's the case, I'd like to update `TpuGkeTask` to be `TpuXpkTask` to be specific as well.

configs/vm_resource.py
```diff
@@ -32,3 +32,20 @@ class RuntimeVersion(enum.Enum):
...
+
+# TODO(ranran): update the cluster name once quota is approved (b/311073979).
+class ClusterName(enum.Enum):
+  V4_CLUSTER = ""
+  V5E_CLUSTER = "ran-xpk-test-zone"
+
+
+# TODO(ranran): update the cluster name once quota is approved (b/311073979).
+class ClusterConfig(enum.Enum):
+  V4_CONFIG = "v4_cluster_config"
```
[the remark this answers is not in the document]
RissyRan: Thanks! My understanding the way xpk works is to use all of TPUs in a cluster, and it cannot use partial of them. i.e. we have a cluster in region us-central-2, and have TPU v4 and v5e in different pool ([example](https://screenshot.googleplex.com/BFNC7bpjyBFPzFP)). But I will need to double check with Victor on this.

dags/benchmark/pytorchxla_torchbench.py
RissyRan: Oh.. I think this is introduced due to the merge a while ago. The head has been changed already, I will update in the next commit.

implementations/utils/xpk.py
```diff
@@ -0,0 +1,106 @@
...
+  """
+
+  cmds = (
+      "set -x",
+      f"gcloud config set project {project_id}",
+      f"gcloud config set compute/zone {zone}",
+      "git clone -b xpk-namespace https://github.com/google/xpk.git /tmp/xpk",
+      "cd /tmp/xpk",
+      (
+          "python3 xpk.py workload create"
```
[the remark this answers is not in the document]
RissyRan: I don't. All examples are provided for bash script here (https://github.com/google/xpk), but I will double check with Victor on this.

implementations/utils/metric.py
RissyRan: I did not find a good way to get the status when running `run_workload` via bash script. There is a bug for xpk to wait till completion (b/306688270). I tried command in that bug to get status of workload, but unfortunately, we will need extra efforts to clean up output (output including extra logs & tables).

implementations/utils/xpk.py
RissyRan: Sounds good!

configs/example/gke_example_config.py
RissyRan: This is a good point. Yes, currently I just want to show an example. Since we are testing models, I was thinking probably it's better to have a hello world E2E example. Do you feel this is a complex one?

implementations/utils/metric.py
RissyRan: One example for status: https://screenshot.googleplex.com/BpDxsaUX6nh6BJV We will need to parse both status and status message to see if this job is failed or successful.

apis/test_config.py
```diff
@@ -78,7 +79,7 @@ class Tpu(Accelerator):
 
   version: str
   cores: int
-  runtime_version: str
+  runtime_version: str = vm_resource.RuntimeVersion.TPU_UBUNTU2204_BASE.value
```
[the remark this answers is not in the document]
RissyRan: Good call! Updated it to `runtime_version: Optional[str] = None`

implementations/utils/xpk.py
RissyRan: Update to something like `from kubernetes import client as kubernetes_client, config as kubernetes_config`

implementations/utils/xpk.py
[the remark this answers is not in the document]
RissyRan: No, unfortunately, there is no lib plan on xpk.

implementations/utils/xpk.py
RissyRan: Discussed offline. It seems SubprocessHook is not compatible with @task.docker and @task.kubernetes. Not sure if this is Composer specific.

implementations/utils/metric.py
will-cromar: Got it. Let's revisit this if/when XPK adds a Python API and see if there's a better way to plumb the status.

PR #33, opened 2023-12-04
Pull the latest submodule commit

RissyRan:
Change-Id: Ia43089a8224e7f258eb1a9b44876ebd8f0c2b20f

# Description

Pull the latest submodule commit and add Jsonnet configs:
* address issues that missing [configs](https://screenshot.googleplex.com/4n9jAxzy5JpnPCA)
* convenience for future test config debugging before deprecation
* configs will be auto pushed to GCS bucket belongs with the composer env

# Tests

```
git submodule init
git submodule update
cd ml-testing-accelerators/
git checkout master
git pull
cd ..
git add ml-testing-accelerators/
git commit -m "Pull the latest commit from submodule"
```

Re-generate the configs:

```
bash scripts/gen-configs.sh
```

# Checklist


.gitignore
```diff
@@ -7,6 +7,3 @@ __pycache__
 # .tfstate files
 *.tfstate
 *.tfstate.*
-
-# Compiled JSonnet templates
-configs/jsonnet/*
```
will-cromar, line 12: We should ignore these so we don't pollute the diffs. Most of the tests generated from the other repository are not used here.

will-cromar (commented):
As a quick fix for the failing DAG check, you can add a `-1vm` suffix here until the name is fixed upstream: https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/bdb22c31138ff1f81506ea3d123c13864c4a8a83/dags/xlml/pytorchxla_llama.py#L37-L46

.gitignore
will-cromar: Oh, I see what's wrong. The auto-deployment pipeline doesn't build the tests. Let me see if there's a snippet you can copy/paste to build the JSonnet templates as part of the deployment pipeline. We build the templates [dynamically for GHA](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/bdb22c31138ff1f81506ea3d123c13864c4a8a83/.github/workflows/dag-check.yml#L18-L35), but the environment is a little different for GCB. I can take a look tomorrow if there's no easy fix.

RissyRan: Thanks for the fix! Removed Jsonnet configs in this PR.

will-cromar (approved):
Thanks!

PR #34, opened 2023-12-04
Build JSonnet files as part of the postsubmit

will-cromar:
# Description

Build the configs from the other repository before uploading configs to the DAGs bucket. Updated from an old snippet we used to use when we worked out of Cloud Source Repositories.

# Tests

Test run: http://shortn/_01mv8E9kv8

We can actually still run the postsubmit in less than a minute (vs ~3 minutes before) by using `slim`/`alpine` images to cut out overhead from image pulls.

# Checklist

### DIFF
--- a/pipeline/auto-push.cloudbuild.yml
+++ b/pipeline/auto-push.cloudbuild.yml
@@ -1,6 +1,22 @@
 steps:
-  - name: google/cloud-sdk
-    args:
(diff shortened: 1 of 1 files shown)


RissyRan (approved):
Looks great! Thanks!

I will remove Jsonnet configs from my PR, but update the submodule only.

PR #45, opened 2023-12-09
add GPU support for pytorch torchbench benchmarking

zpcore:
# Description

- Add `vm_api.gpu_api` to mimic `tpu_api`. `gpu_api` supports resource creation and remote execution using fabric.
- Fix the torchbench TPU run config to support nightly torch/torchaudio/torchvision build.
- Add the torchbench GPU run for nightly test.
- Add platform support for  GPU V100/A100/L4/H100 and TPU V5E/V5P/V4. H100 and A100 are waiting for the resource capacity to be ready.

# Tests

airflow test link [here](https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla-torchbench/grid).

Please wait for the test to complete.

# Checklist


zpcore:
The workers seem dead when I launch 10 runs in parallel. Is there a resource limitation?

apis/test_config.py
```diff
@@ -90,6 +89,31 @@ def name(self):
     return f'v{self.version}-{self.cores}'
 
 
+# We may want to use GKE in the future.
```
RissyRan, line 92: Probably update it to a TODO comment? i.e. add GKE support for GPU. If you don't know who will implement the feature, feel free to assign it to me.

apis/test_config.py
```diff
@@ -90,6 +89,31 @@ def name(self):
...
+  """
+
+  machine_type: str
+  image_family: str
+  count: str
+  accelerator_type: str
+  runtime_version: str
+
+  @property
+  def name(self):
```
RissyRan, line 112: Could you help me understand more here? could we directly call `self.accelerator_type` if there is no extra processing step?

apis/test_config.py
```diff
@@ -262,8 +313,7 @@ def from_pytorch(test_name: str, reserved_tpu: bool = True):
     return JSonnetTpuVmTest._from_json_helper(
         test,
         setup=test['tpuSettings']['tpuVmPytorchSetup']
-        # HACK: Extra setup assumes a new shell in home directory
-        + '\ncd ~\n' + test['tpuSettings']['tpuVmExtraSetup'],
+        + test['tpuSettings']['tpuVmExtraSetup'],
```
RissyRan, line 316: I assume you have tested it on Airflow?

configs/benchmark/pytorch/pytorchxla_torchbench_config.py
```diff
@@ -119,3 +105,131 @@ def get_torchbench_config(
       task_gcp_config=job_gcp_config,
       task_metric_config=job_metric_config,
   )
+
+
+# Below is the setup for torchbench GPU run.
```
RissyRan, line 110: Nit: I think we could remove this comment as method name `set_up_torchbench_gpu` explains it already. Same comment for other methods under this situation.

implementations/utils/gpu.py
```diff
@@ -0,0 +1,113 @@
...
+from absl import logging
+import airflow
+from airflow.decorators import task, task_group
+from airflow.utils.task_group import TaskGroup
+from apis import gcp_config, test_config
+import fabric
+import google.api_core.exceptions
+import google.auth
+# There is no GPU api to reserve for resources. So we use gcloud cmdline to reserve resouces.
+# import google.cloud.tpu_v2alpha1 as tpu_api
```
RissyRan, line 30: Remove?

implementations/utils/gpu.py
```diff
@@ -0,0 +1,113 @@
...
+from implementations.utils import ssh
+from implementations.utils.vm_api import gpu_api
+import paramiko
+
+
+@task
+def generate_gpu_name(base_gpu_name: str) -> str:
+  # note: GPU vm name need to match regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)', while TPU vm allows '_'.
+  # return f'{base_gpu_name}-{str(uuid.uuid4())}'.replace('_', '-')
+  return f'gpu-{str(uuid.uuid4())}'
```
RissyRan, line 41: I see the name in the comment is `return f'{base_gpu_name}-{str(uuid.uuid4())}'.replace('_', '-')`. Do we plan to use that or `f'gpu-{str(uuid.uuid4())}'`? I prefer the former one if possible.

apis/task.py
```diff
@@ -271,11 +268,107 @@ class GpuTask(BaseTask):
...
+
+    Returns:
+      A task group with the following tasks chained: provision, run_model,
+      post_process and clean_up.
+    """
+    # piz: We skip the queued resource for GPU for now since there is no queued resource command for GPU.
+    with TaskGroup(
+        group_id=self.task_test_config.benchmark_id, prefix_group_id=True
+    ) as group:
+      provision, queued_resource, ssh_keys = self.provision()
```
RissyRan, line 292: If tehre is no Queued Resource for GPU, we could update those `queued_resource` to something like `resource`. Please check other variable names in this file.

implementations/utils/vm_api.py
```diff
@@ -0,0 +1,195 @@
...
+          f'--image-family={self.image_family}',
+          f'--image-project={self.image_project}',
+          f'--project={self.project}',
+          f'--accelerator=type={self.accelerator_type},count={self.count}',
+          '--boot-disk-size=500GB',
+          f'--metadata="install-nvidia-driver=True,proxy-mode=project_editors,ssh-keys=cloud-ml-auto-solutions:{self.ssh_keys.public}"',
+          '--scopes=https://www.googleapis.com/auth/cloud-platform',
+          '--on-host-maintenance=TERMINATE',
+          '--instance-termination-action=DELETE',
+          '--max-run-duration=60m',  # TODO(piz): for prototype purpose
```
RissyRan, line 130: Could we pass in the timeout limit here?

implementations/utils/vm_api.py
```diff
@@ -0,0 +1,195 @@
...
+import subprocess
+from typing import Optional, Sequence, TextIO
+from absl import logging
+
+
+class RunError(Exception):
+  pass
+
+
+class gpu_api:
```
RissyRan, line 24: In my option, use BashOperator to implement `build()` will be simpler. What's your thoughts? @will-cromar

apis/task.py
```diff
@@ -271,11 +268,107 @@ class GpuTask(BaseTask):
   """This is a class to set up tasks for GPU.
```
RissyRan, line 268: Currently, TPU tasks are named as `TpuQueuedResourceTask` and `TpuXpkTask`. I did not think of a better name here. Your thoughts?

RissyRan (commented):
Thanks Pei for adding GPU tasks! @will-cromar, could you also take a look when you get a chance?

apis/test_config.py
zpcore: I leave the name there in case we want something like `nvidia-tesla-v100-4` or just `v100` instead of `nvidia-tesla-v100` in the future. WDYT?

apis/test_config.py
zpcore: I don't remember why I did that, let me revert this back and redo a test.

configs/benchmark/pytorch/pytorchxla_torchbench_config.py
zpcore: Done

implementations/utils/gpu.py
zpcore: Done

implementations/utils/gpu.py
zpcore: Thanks, I tried the former one before. However due to the `base_gpu_name` is too long, the full string will be long than expected (there is a requirement that some potion of the string should be less than 61 as in the regex mentioned).  Any ideas for a better naming?

RissyRan: I see. If that's the case, could you add a comment indicates there is a limitation for len of characters?

apis/test_config.py
RissyRan: I see. Probably we could add a TODO comment, or we could directly call `self.accelerator_type` and update it in the future when we are clear on names?

implementations/utils/gpu.py
zpcore: Comment added.

apis/task.py
zpcore: I changed it to `GpuCreateResourceTask`. If we have a better name, I can make the update.

implementations/utils/vm_api.py
zpcore: Thanks, I didn't know we have `BashOperator`  to use. I can submit a separate PR to change to `BashOperator`.

apis/task.py
zpcore: Sounds good, updated

apis/test_config.py
```diff
@@ -90,6 +89,31 @@ def name(self):
     return f'v{self.version}-{self.cores}'
 
 
+# TODO(ranran): We may want to use GKE in the future.
```
RissyRan, line 92: Sorry for the confusion if that happens. Could we move this comment to `def name(self):`? so that I know we need to update it.

apis/test_config.py
```diff
@@ -90,6 +89,31 @@ def name(self):
...
+@attrs.define
+class Gpu(Accelerator):
+  """Represents a single Cloud GPU instance.
+
+  Attributes:
+    machine_type: The host type of the GPU. E.g., `a2-highgpu-1g`.
+    image_family: Family of the image.
+    count: number of the GPU devices.
```
RissyRan, line 100: Nit: Capital N.

apis/test_config.py
```diff
@@ -155,6 +179,34 @@ def test_script(self) -> str:
...
+  """Test config that runs on a single Cloud GPU VM instance.
+
+  Attributes:
+    test_name: Unique name for this test/model.
+    set_up_cmds: List of commands to run once when GPU is created.
+    run_model_cmds: List of commands to run the model under test.
+  """
+
+  test_name: str
+  vm_duration: str
```
RissyRan, line 193: What's the difference between `vm_duration` and `time_out_in_min`? Can we directly use that?

Also miss the comment in the attributes if they are difference.

apis/test_config.py
```diff
@@ -264,6 +316,7 @@ def from_pytorch(test_name: str, reserved_tpu: bool = True):
         setup=test['tpuSettings']['tpuVmPytorchSetup']
         # HACK: Extra setup assumes a new shell in home directory
         + '\ncd ~\n' + test['tpuSettings']['tpuVmExtraSetup'],
+        # + test['tpuSettings']['tpuVmExtraSetup'],
```
RissyRan, line 319: What's this for in the comment?

configs/benchmark/pytorch/pytorchxla_torchbench_config.py
RissyRan: I still see this comment?

implementations/utils/gpu.py
```diff
@@ -0,0 +1,111 @@
...
+    image_project: str,
+    image_family: str,
+    accelerator,
+    vm_duration,
+    gcp: gcp_config.GCPConfig,
+    ssh_keys: airflow.XComArg,
+) -> Tuple[TaskGroup, airflow.XComArg]:
+  """Request a resource and wait until the nodes are created.
+
+  Args:
```
RissyRan, line 54: Missing some Args in the comment. Also, wondering the reason that `create_resource_request` instead of `create_resource` method? We don't need to wrap it in a task group if there is only 1 task.

implementations/utils/vm_api.py
RissyRan: I think it will be clearer if we use BashOperator for that build(). A few examples here: https://airflow.apache.org/docs/apache-airflow/stable/howto/operator/bash.html

implementations/utils/gpu.py
```diff
@@ -0,0 +1,111 @@
...
+        accelerator,
+        vm_duration,
+        ssh_keys,
+    )
+    logging.info(
+        'CREATE TPU RESOURCE: create_node_request. IP:' f' {gpu_api.node.ip_address}'
+    )
+    return gpu_api.node.ip_address
+
+  with TaskGroup(group_id='create_queued_resource') as tg:
```
RissyRan, line 83: Wondering why do we need to have a TaskGroup here with one `create_resource_request` task inside?

RissyRan (commented):
Thanks Pei! I am tying to understand better on codes. Also, please check your files to ensure pre-submit checks are all green.

implementations/utils/vm_api.py
RissyRan: Let's also discuss it offline.

implementations/utils/gpu.py
```diff
@@ -0,0 +1,111 @@
...
+    qualified_name = create_resource_request(gpu_name, ssh_keys)
+    # It takes time for the SSH key to be propagated to the instance.
+    # We may fail to connect in the first trial.
+  return tg, qualified_name
+
+
+@task
+def ssh_gpu(ip_address: str, cmds: Iterable[str], ssh_keys: ssh.SshKeys) -> None:
```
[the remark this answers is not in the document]
zpcore: SGTM! Also fixed the comments related to TPU.

apis/test_config.py
zpcore: Removed the comment.

implementations/utils/gpu.py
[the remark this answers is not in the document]
zpcore: Oh, just noticed that tpu used ssh_tpu() in https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/98bdf439ae5e7b8ad7505e011e2352447039709f/implementations/utils/tpu.py#L221. I think it may be better to follow the same rule. 

Also `ssh_tpu` can prevent messing ssh in `ssh_keys: ssh.SshKeys` with the local ssh_tpu() function.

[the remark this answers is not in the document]
zpcore: maybe change both `ssh_tpu` and `ssh_gpu` to `ssh_host`, WDYT?

implementations/utils/gpu.py
zpcore: Removed the wrap.

implementations/utils/gpu.py
zpcore: Removed the wrap.

apis/test_config.py
zpcore: There are several `def name(self):`, which `def name(self):` do you mean here?

apis/task.py
```diff
@@ -265,15 +262,112 @@ def post_process(self) -> DAGNode:
...
+  task_test_config: test_config.TestConfig[test_config.Gpu]
+  task_gcp_config: gcp_config.GCPConfig
+  task_metric_config: Optional[metric_config.MetricConfig] = None
+
+  def run(self) -> DAGNode:
+    """Run a test job.
+
+    Returns:
+      A task group with the following tasks chained: provision, run_model,
+      post_process and clean_up.
```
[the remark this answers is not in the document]
zpcore: Removed the `clean_up`.

configs/benchmark/pytorch/pytorchxla_torchbench_config.py
```diff
@@ -119,3 +107,132 @@ def get_torchbench_config(
...
+def set_up_torchbench_gpu(model_name: str = "") -> Tuple[str]:
+  """Common set up for TorchBench."""
+  # TODO(piz): There is issue with driver install through fabric.
```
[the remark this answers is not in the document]
zpcore: The issue is installing GPU driver through fabric. The installation will eventually fail due to cuda is in use. I tried many ways to kill the cuda related process before the installation but still doesn't work. One assumption I think is that fabric connects with the host and enables the x server, which prevents the driver installation since the x server is using the cuda.

configs/benchmark/pytorch/pytorchxla_torchbench_config.py
```diff
@@ -30,40 +30,28 @@ def set_up_torchbench(model_name: str = "") -> Tuple[str]:
       "sudo apt install -y libopenblas-base",
       "sudo apt install -y libsndfile-dev",
       "sudo apt-get install libgl1 -y",
+      "sudo chmod 777 /usr/local/lib/python3.10/dist-packages/",
```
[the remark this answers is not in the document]
zpcore: Yes, that should also work. I can update and do the test.

One concern for using `--user` is that the installation path is not in the default `$PATH`, and this can cause the installation complain.

implementations/utils/vm_api.py
```diff
@@ -0,0 +1,199 @@
...
```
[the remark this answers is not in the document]
zpcore: Updated to use the gce API.

apis/task.py
```diff
@@ -265,15 +262,134 @@ def post_process(self) -> DAGNode:
...
+  task_metric_config: Optional[metric_config.MetricConfig] = None
+
+  def run(self) -> DAGNode:
+    """Run a test job.
+
+    Returns:
+      A task group with the following tasks chained: provision, run_model,
+      post_process.
+    """
+    # piz: We skip the queued resource for GPU for now since there is no queued resource command for GPU.
```
miladm, line 286: what happens if the resource was unavailable? does it mean we have a higher chance of resource drop without queued resource?

configs/benchmark/pytorch/pytorchxla_torchbench_config.py
```diff
@@ -119,3 +125,142 @@ def get_torchbench_config(
       task_gcp_config=job_gcp_config,
       task_metric_config=job_metric_config,
   )
+
+
+# Below is the setup for torchbench GPU run.
+def set_up_torchbench_gpu(model_name: str = "") -> Tuple[str]:
```
miladm, line 131: let's confirm the same code can run on AMD GPUs in the future please.

dags/benchmark/pytorchxla_torchbench.py
```diff
@@ -19,31 +19,111 @@
...
 
-# Schudule the job to run every day at 5:00PM UTC.
-SCHEDULED_TIME = "0 17 * * *" if composer_env.is_prod_env() else None
 
 with models.DAG(
     dag_id="pytorchxla-torchbench",
     schedule=SCHEDULED_TIME,
     tags=["pytorchxla", "nightly", "torchbench"],
-    start_date=datetime.datetime(2023, 8, 29),
+    start_date=datetime.datetime(2024, 1, 1),
```
miladm, line 34: nit question: why is the date hardcoded?

apis/test_config.py
RissyRan: For this comment, could you help remind me if we want to move your GPU implementation to use GKE? or we need a separate implementation like `TpuGkeTest`? 

If it's former, it makes sense to add the comment here, but you may be the assignee? If the later, it will be a new feature request. I don't think you need to add a TODO here. Let me know your thoughts!

configs/benchmark/pytorch/pytorchxla_torchbench_config.py
```diff
@@ -119,3 +107,132 @@ def get_torchbench_config(
...
+      (
+          "sudo docker pull"
+          " us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.8_cuda_12.1"
+      ),
+      (
+          "sudo docker run --gpus all -it -d"
+          " us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.8_cuda_12.1"
+      ),
+      (
+          "sudo docker exec -i $(sudo docker ps | awk 'NR==2 { print $1 }')"
```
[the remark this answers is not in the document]
zpcore: Hmm, tried to the `... --entrypoint bash gcr.io/image/url -c 'docker_cmds'` but didn't work as expected. Below is the command I tried: 
```
sudo docker run --gpus all -i -d --network host us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.8_cuda_12.1 --entrypoint /bin/bash us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.8_cuda_12.1 -c 'pip install --user --pre torchvision torchaudio -i &&  https://download.pytorch.org/whl/nightly/cu121'"
```
Do you have any hint on how to specify the command using `--entrypoint`?

dags/benchmark/pytorchxla_torchbench.py
RissyRan: This start date does not quite matter, if we set up [catchup](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dag-run.html#catchup) as False (so no backfill test will run). It indicates when we initialize the scheduled tests.

dags/benchmark/pytorchxla_torchbench.py
```diff
@@ -19,31 +19,111 @@
...
-        model_name=model,
-        time_out_in_min=60,
-        extraFlags=" ".join(torchbench_extra_flags),
-    ).run()
+  model = "BERT_pytorch"
+  model = "all"
+  torchbench_extra_flags = [f"--filter={model}"]
+  # Running on V4-8:
+  config.get_torchbench_tpu_config(
+      tpu_version="4",
```
RissyRan, line 42: Nit: can we reuse version here to avoid potential typo? https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/configs/vm_resource.py#L43

configs/benchmark/pytorch/pytorchxla_torchbench_config.py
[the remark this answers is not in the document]
zpcore: But I think it is better to seperate the `docker run` and `docker exec` since we need to copy experiment data out from the docker with `docker exec` during postprocess stage.

dags/benchmark/pytorchxla_torchbench.py
```diff
@@ -19,31 +19,111 @@
...
+      network=V5_NETWORKS,
+      subnetwork=V5E_SUBNETWORKS,
+      time_out_in_min=1600,
+      model_name=model,
+      extraFlags=" ".join(torchbench_extra_flags),
+  ).run()
+
+  # Running on V100 GPU
+  config.get_torchbench_gpu_config(
+      machine_type="n1-standard-32",
```
RissyRan, line 81: Can you add those GPU types as enum in `vm_resource.py`? like TPU types? https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/configs/vm_resource.py#L43

configs/benchmark/pytorch/pytorchxla_torchbench_config.py
```diff
@@ -119,3 +107,132 @@ def get_torchbench_config(
...
       task_metric_config=job_metric_config,
   )
+
+
+# Below is the setup for torchbench GPU run.
+def set_up_torchbench_gpu(model_name: str = "") -> Tuple[str]:
+  """Common set up for TorchBench."""
+  # TODO(piz): There is issue with driver install through fabric.
+  # Currently we use pre-installed driver to avoid driver reinstall.
+  nvidia_install_clean = (
```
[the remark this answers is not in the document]
zpcore: Nice, the installation works. Currently the cuda version is 12.0. I am using it by default for now, but it is better to find the 12.1 version.

configs/benchmark/pytorch/pytorchxla_torchbench_config.py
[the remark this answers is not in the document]
zpcore: Fixed based on comment: https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/45#discussion_r1424694127.

dags/benchmark/pytorchxla_torchbench.py
zpcore: Added those configuration classes and updated the config.

dags/benchmark/pytorchxla_torchbench.py
zpcore: Updated along with https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/45#discussion_r1445409968.

apis/task.py
zpcore: Yes, the request will be dropped after 3 attempts if resource is still unavailable. We should find a solution for queued resource for GPU in the future.

configs/benchmark/pytorch/pytorchxla_torchbench_config.py
zpcore: Will do the test once we have AMD GPU resource available.

implementations/utils/vm_api.py
```diff
@@ -0,0 +1,377 @@
...
+      msg = str(e)
+      if stderr and stderr.seekable() and stderr.readable():
+        stderr.flush()
+        stderr.seek(0)
+        lines = stderr.readlines()[-10:]
+        msg += f'.\n    Last {len(lines)} lines of standard error:\n      '
+        msg += '      '.join(lines)
+      raise RunError(msg) from e
+
+  #   def run(vm_name, zone, project, cmd):
```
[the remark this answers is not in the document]
zpcore: The whole file is removed. All the functions have been changed into gcp api and moved into gpu.py.

implementations/utils/vm_api.py
```diff
@@ -0,0 +1,377 @@
...
+    instance.deletion_protection = True
+
+  # Prepare the request to insert an instance.
+  request = compute_v1.InsertInstanceRequest()
+  request.zone = zone
+  request.project = project_id
+  request.instance_resource = instance
+
+  # Wait for the create operation to complete.
+  print(f'Creating the {instance_name} instance in {zone}...')
```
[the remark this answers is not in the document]
zpcore: The whole file is removed. All the functions have been changed into gcp api and moved into gpu.py.

apis/task.py
```diff
@@ -265,15 +262,134 @@ def post_process(self) -> DAGNode:
...
+    )
+
+  def post_process(self) -> DAGNode:
+    """Process metrics and metadata, and insert them into BigQuery tables.
+
+    Returns:
+      A DAG node that executes the post process.
+    """
+    with TaskGroup(group_id="post_process") as group:
+      process_id = metric.generate_process_id.override(retries=1)()
```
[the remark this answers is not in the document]
zpcore: Done!

apis/task.py
```diff
@@ -265,15 +262,134 @@ def post_process(self) -> DAGNode:
...
+      post_process.
+    """
+    # piz: We skip the queued resource for GPU for now since there is no queued resource command for GPU.
+    with TaskGroup(
+        group_id=self.task_test_config.benchmark_id, prefix_group_id=True
+    ) as group:
+      provision, ip_address, instance_name, ssh_keys = self.provision()
+      run_model = self.run_model(ip_address, ssh_keys)
+      post_process = self.post_process()
+      clean_up_process = self.clean_up(
```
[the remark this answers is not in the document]
zpcore: Yes, I setup the `instance_termination_action=DELETE` action when create the VM and meanwhile add the operation to delete the resource when everything is done.  Out of curiosity, will `instance_termination_action=DELETE` automatically delete the instance immediately when we finish the experiment run?

configs/vm_resource.py
```diff
@@ -29,6 +29,14 @@ class Project(enum.Enum):
   TPU_PROD_ENV_AUTOMATED = "tpu-prod-env-automated"
 
 
+class ImageProject(enum.Enum):
+  DEEP_LEARNING_PLATFORM_RELEASE = ("deeplearning-platform-release",)
```
RissyRan, line 33: Is there special requirement for Image project so that we need tuple of strings?

implementations/utils/gpu.py
```diff
@@ -0,0 +1,457 @@
...
+
+  Returns:
+    The output as a string.
+  """
+  images_client = compute_v1.ImagesClient()
+  # Listing only non-deprecated images to reduce the size of the reply.
+  images_list_request = compute_v1.ListImagesRequest(
+      project=project,
+      max_results=100,
+      # filter="deprecated.state != DEPRECATED",
```
RissyRan, line 51: Nit: what's this comment?

implementations/utils/gpu.py
```diff
@@ -0,0 +1,457 @@
...
+from implementations.utils import ssh
+import io
+import paramiko
+import re
+import sys
+from typing import Iterable, Any
+import uuid
+
+
+def print_images_list(project: str) -> str:
```
RissyRan, line 36: Could you help me understand what's the purpose of printing the image list? Is this for debugging purpose? Thanks!

implementations/utils/gpu.py
```diff
@@ -0,0 +1,457 @@
...
+    instance.deletion_protection = True
+
+  # Prepare the request to insert an instance.
+  request = compute_v1.InsertInstanceRequest()
+  request.zone = zone
+  request.project = project_id
+  request.instance_resource = instance
+
+  # Wait for the create operation to complete.
+  print(f"Creating the {instance_name} instance in {zone}...")
```
RissyRan, line 252: Nit: logging.info instead of print. Similar for other occurrences if apply.

implementations/utils/gpu.py
```diff
@@ -0,0 +1,457 @@
...
+          named "default", which is created automatically for each project.
+      subnetwork_link: name of the subnetwork you want the new instance to use.
+          This value uses the following format:
+          "regions/{region}/subnetworks/{subnetwork_name}"
+
+  Returns:
+      Instance object.
+  """
+  newest_debian = get_image_from_family(project="debian-cloud", family="debian-10")
+  disk_type = f"zones/{zone}/diskTypes/pd-standard"
```
RissyRan, line 337: Should this be a variable to pass in the disk type? Or it's not necessary here.

implementations/utils/gpu.py
```diff
@@ -0,0 +1,457 @@
...
+  return result
+
+
+def create_metadata(key_val: dict[str, str]) -> compute_v1.Metadata:
+  metadata = compute_v1.Metadata()
+  metadata.items = [{"key": key, "value": val} for key, val in key_val.items()]
+  return metadata
+
+
+def create_with_subnet(
```
RissyRan, line 316: Probably I missed it somewhere. Could you remind me where we used this?

RissyRan (commented):
Thanks Pei for the quick turn around! I have some questions for my own understanding about GPU. Others seem good to me.

configs/vm_resource.py
zpcore: Oh, thanks for catching the typo. That's why all the GPU tests failed.

apis/test_config.py
zpcore: Prefer remove the TODO item. Unless we have a solid motivation to move to GKE.

implementations/utils/gpu.py
zpcore: This was used to fetch the image from the public repo. Since we use predefined image name, there is no need for the filter anymore. Let me remove it.

implementations/utils/gpu.py
zpcore: Remove the function since we use the predefined image.

configs/benchmark/pytorch/pytorchxla_torchbench_config.py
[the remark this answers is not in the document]
zpcore: Updated to use docker name instead of `awk`.

implementations/utils/tpu.py
```diff
@@ -223,7 +223,7 @@ def wait_for_queued_resource_deletion(op_name: Optional[str]):
 
 
 @task
-def ssh_tpu(
+def ssh_host(
```
will-cromar, line 226: nit: keep the name `ssh_tpu` or similar. The semantics are different than `gpu.ssh_host`, since this will SSH to _multiple_ hosts in a TPU resource.

implementations/utils/gpu.py
```diff
@@ -0,0 +1,394 @@
...
+
+  operation = instance_client.insert(request=request)
+
+  wait_for_extended_operation(operation, "instance creation")
+
+  logging.info(f"Instance {instance_name} created.")
+  return instance_client.get(project=project_id, zone=zone, instance=instance_name)
+
+
+def wait_for_extended_operation(
```
will-cromar, line 234: nit: this can be a `task.sensor` like `wait_for_queued_resource`

implementations/utils/gpu.py
```diff
@@ -0,0 +1,394 @@
...
+  initialize_params.disk_type = disk_type
+  boot_disk.initialize_params = initialize_params
+  # Remember to set auto_delete to True if you want the disk to be deleted when you delete
+  # your VM instance.
+  boot_disk.auto_delete = auto_delete
+  boot_disk.boot = boot
+  return boot_disk
+
+
+def create_instance(
```
will-cromar, line 91: What's the benefit of separating this from `create_resource`? IMO it's more readable to inline the parts you need there, like `compute_v1.Instance(name=instance_name, ...)`

zpcore: My thinking is that `create_resource` is the function that takes input related to our automation setup like customized metadata, while `create_instance` is a more general api for creating a VM.

configs/vm_resource.py
```diff
@@ -29,6 +29,14 @@ class Project(enum.Enum):
   TPU_PROD_ENV_AUTOMATED = "tpu-prod-env-automated"
 
 
+class ImageProject(enum.Enum):
+  DEEP_LEARNING_PLATFORM_RELEASE = "deeplearning-platform-release"
```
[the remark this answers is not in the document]
zpcore: My concern is that if in the future we use VM that doesn't require the reinstallation of the drivers, removing he image may cause issue in that case. At this time I added `--force` in `sudo python3 install_gpu_driver.py --force` to force the installation.

I think we can keep it here since it won't do any harm in case driver is not reinstalled in the future. WDYT?

apis/task.py
```diff
@@ -256,15 +254,135 @@ def post_process(self) -> DAGNode:
...
+      resource: an XCom value for the qualified instance name.
+      project_id: project of the instance.
+      zone: zone of the instance.
+    Returns:
+      A DAG node that deletes the queued resource and its owned nodes.
+
+    Raises:
+      AirflowTaskTimeout: An error occurs when execution_timeout is breached.
+    """
+    with TaskGroup(group_id="clean_up") as group:
```
RissyRan, line 384: We don't need a group here right?

dags/benchmark/pytorchxla_torchbench.py
```diff
@@ -14,37 +14,113 @@
 
 """A DAG to run all TorchBench tests with nightly version."""
 
-import datetime
 from airflow import models
+import datetime
 from configs import composer_env
 from configs.benchmark.pytorch import pytorchxla_torchbench_config as config
-from configs.vm_resource import TpuVersion, Zone
+from configs.vm_resource import *
```
RissyRan, line 21: I think wildcard import this is not recommended? As we don't know where it comes from.

RissyRan (commented):
Could you attach your latest GPU test link? Thanks!

implementations/utils/gpu.py
zpcore: After adding `@task.sensor(poke_interval=60, timeout=3600, mode="reschedule")` for the function, the GPU VM is created correctly but the task is still marked to retry.  And the second time we retry build, it will error out complaining the tpu name already exists.

zpcore:
> Could you attach your latest GPU test link? Thanks!

Here is the [link](https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla-torchbench/grid?dag_run_id=manual__2024-01-11T01%3A47%3A37.509447%2B00%3A00&tab=graph&task_id=torchbench_all-nvidia-l4). Currently the run models failed to run all the model due to the benchmarking coding issue, which is not associated with the DAG setup.

implementations/utils/gpu.py
RissyRan: Have you updated the return value? The return value of the sensor needs to be a bool. https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/implementations/utils/tpu.py#L118

implementations/utils/gpu.py
```diff
@@ -0,0 +1,386 @@
...
+          "auth_strategy": paramiko.auth_strategy.InMemoryPrivateKey(
+              "cloud-ml-auto-solutions", pkey
+          )
+      },
+  )
+  ssh_group.run(cmds)
+
+
+@task(trigger_rule="all_done")
+def delete_resource(instance_name: str, project_id: str, zone: str) -> str:
```
RissyRan, line 377: I agree with Will. It's better we have a func called delete_resource_request, and sensor called wait_for_resource_deletion to be consistent with existing implementation. https://screenshot.googleplex.com/8uMfXiEXzoFqv62

apis/task.py
```diff
@@ -256,15 +254,133 @@ def post_process(self) -> DAGNode:
...
+      resource: an XCom value for the qualified instance name.
+      project_id: project of the instance.
+      zone: zone of the instance.
+    Returns:
+      A DAG node that deletes the queued resource and its owned nodes.
+
+    Raises:
+      AirflowTaskTimeout: An error occurs when execution_timeout is breached.
+    """
+    return gpu.delete_resource.override(task_id="release_resource")(
```
RissyRan, line 384: Similar comment. We should name it as `clean_up` task group, and then has a task delete_resource_request, and sensor called wait_for_resource_deletion, instead of wrapping them together: https://screenshot.googleplex.com/8uMfXiEXzoFqv62

RissyRan (commented):
Thanks Pei! You may want to ensure the [DAG check pass](https://github.com/GoogleCloudPlatform/ml-auto-solutions/actions/runs/7495795813/job/20406563237?pr=45), currently it failing at `ImportError: cannot import name 'compute_v1' from 'google.cloud' (unknown location)`.

You may want to add proper dependency into this [file](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/.github/requirements.txt). You cannot merge the code unless all checks pass. 

Packages may be already in the [version list](https://cloud.google.com/composer/docs/concepts/versioning/composer-versions), but not in the check.

implementations/utils/gpu.py
RissyRan: Based on the current workflow, what will be the use cases for:

1) we set `instance_termination_action` as stop, and then clean it up at the end
2) we set `instance_termination_action` as delete, and then clean it up at the end

Do we need this clean up? or do we need to add `instance_termination_action` param. Thanks?

implementations/utils/gpu.py
zpcore: I got what you mean here. Will remove the `wait_for_extended_operation` and use sensor instead for resource creation in a separate PR. I have updated the solution to resource deletion.

implementations/utils/gpu.py
zpcore: Updated to use STOP as the termination action since there is resource_deletion in the cleanup stage.

apis/task.py
zpcore: Thanks for the point. I have changed to task_sensor instead of the `wait_for_extended_operation` and it works well. WILL has a similar feedback regarding the resource_creation. I will make the update for that in a separate PR.

apis/test_config.py
zpcore: `vm_duration` has been removed in later changes.

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
```diff
@@ -16,77 +16,79 @@
 
 from typing import Tuple
 from xlml.apis import gcp_config, metric_config, task, test_config
+import dags.vm_resource as resource
 from dags import gcs_bucket, test_owner
-from dags.vm_resource import TpuVersion, Project, RuntimeVersion
 
 
 # TODO(ranran or PyTroch/XLA team): this is an example for benchmark test with hardcode compatible versions,
```
RissyRan, line 23: Could you update this TODO to you, as you are the test owner now?

dags/pytorch_xla/pytorchxla_torchbench.py
```diff
@@ -14,37 +14,113 @@
...
+
+  # Running on V5P
+  config.get_torchbench_tpu_config(
+      tpu_version=resource.TpuVersion.V5P,
+      tpu_cores=8,
+      project=resource.Project.TPU_PROD_ENV_AUTOMATED,
+      tpu_zone=resource.Zone.US_EAST5_A,
+      runtime_version=resource.RuntimeVersion.V2_ALPHA_TPUV5,
+      network=resource.V5_NETWORKS,
+      subnetwork=resource.V5E_SUBNETWORKS,
```
RissyRan, line 56: This should be `V5P_SUBNETWORKS`

RissyRan (commented):
LGTM! Could you pass me your latest test UI link? Thanks!

dags/pytorch_xla/pytorchxla_torchbench.py
zpcore: V5E network still works on v5p platforms as I just tested, but I will update to v5p network to be consistent.

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
zpcore: Haha, sure. I think we can remove the TODO already.

zpcore:
Here is the run with TPU and GPU: https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla-torchbench/grid?dag_run_id=manual__2024-01-19T23%3A29%3A55.062940%2B00%3A00&tab=graph

RissyRan (approved):
LGTM! Please ensure to address other reviewers' comments before the merge! Thanks Pei!

zpcore:
Check in this for now since we are working on the following pending features. Please let me know if there is future improvement feedbacks. Thanks

PR #47, opened 2023-12-12
Enable xpk with multislice tests

RissyRan:
# Description

Enable Multislice ResNet test with xpk.
* Update helper `xpk_example_config.py` to pass `num_slices` and `cluster_project`
* Add `xpk.Dockerfile` to example config folder
* Move `cluster_project` and `cluster_name` to GCP config 
* Other small refactor

# Tests

Upload to Airflow and test in project `cloud-ml-auto-solutions`

Airflow UI: link.

The workload has been added to the queue. Unfortunately, I didn't get a chance to get a successful run, as they are blocked by other jobs.

```
resnet-multi-slice-v4-128-78a3e88d         2023-12-12T01:27:13Z   medium     64               <none>                <none>         Admitted   couldn't assign flavors to pod set slice-job: insufficient unused quota for google.com/tpu in flavor 4xv4-128, 128 more needed   2023-12-12T01:27:13Z
```

# Checklist

### DIFF
--- a/apis/gcp_config.py
+++ b/apis/gcp_config.py
@@ -16,6 +16,7 @@
 
 import dataclasses
 from apis import metric_config
+from typing import Optional
 
 
 @dataclasses.dataclass
@@ -23,11 +24,15 @@ class GCPConfig:
   """This is a class to set up configs of GCP.
 
   Attributes:
-    project_name: The name of a project to run a test job.
+    project_name: The name of a project for Composer env to run a test job.
     zone: The zone to run a test job.
-    dataset_name: The option of dataset for metrics
+    dataset_name: The option of dataset for metrics.
+    cluster_project: The project of a cluster, i.e. a cluster running xpk workloads.
+    cluster_name: The name of a cluster that has provisioned resources.
   """
 
   project_name: str
   zone: str
   dataset_name: metric_config.DatasetOption
+  cluster_project: Optional[str] = None
+  cluster_name: Optional[str] = None
--- a/apis/task.py
+++ b/apis/task.py
@@ -223,9 +223,9 @@ class TpuXpkTask(BaseTask):
       workload_id = xpk.generate_workload_id(self.task_test_config.benchmark_id)
       run_workload = xpk.run_workload(
           task_id="run_workload",
-          project_id=self.task_gcp_config.project_name,
+          cluster_project=self.task_gcp_config.cluster_project,
           zone=self.task_gcp_config.zone,
-          cluster_name=self.task_test_config.cluster_name,
+          cluster_name=self.task_gcp_config.cluster_name,
           benchmark_id=self.task_test_config.benchmark_id,
           workload_id=workload_id,
           docker_image=self.task_test_config.docker_image,
@@ -238,9 +238,9 @@ class TpuXpkTask(BaseTask):
           timeout=self.task_test_config.time_out_in_min * 60,
       )(
           workload_id=workload_id,
-          project_id=self.task_gcp_config.project_name,
+          project_id=self.task_gcp_config.cluster_project,
           region=self.task_gcp_config.zone[:-2],
-          cluster_name=self.task_test_config.cluster_name,
+          cluster_name=self.task_gcp_config.cluster_name,
       )
 
       workload_id >> run_workload >> wait_for_workload_completion
--- a/configs/example/xpk.Dockerfile
+++ b/configs/example/xpk.Dockerfile
@@ -1,0 +1,17 @@
+# Dockerfile for xpk examples in xpk_example_dag.py,
+# and is saved at gcr.io/cloud-ml-auto-solutions/xpk_jax_test:latest.
+FROM python:3.10
+RUN pip install --upgrade pip
+RUN pip install "jax[tpu]" -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
+RUN pip install --no-cache-dir --upgrade clu tensorflow tensorflow-datasets
+RUN git clone https://github.com/google/flax.git /tmp/flax
+RUN pip install --no-cache-dir flax
+RUN curl https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz > /tmp/google-cloud-sdk.tar.gz
+RUN mkdir -p /usr/local/gcloud \
+  && tar -C /usr/local/gcloud -xvf /tmp/google-cloud-sdk.tar.gz \
+  && /usr/local/gcloud/google-cloud-sdk/install.sh
+ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
+RUN gcloud components install kubectl
+RUN gcloud components install gke-gcloud-auth-plugin
+ENV TFDS_DATA_DIR=gs://xl-ml-test-us-central2/tfds-data
+ENV JAX_PLATFORM_NAME=TPU
--- a/deployment/cloud_composer_template.tf
+++ b/deployment/cloud_composer_template.tf
@@ -83,6 +83,14 @@ resource "google_project_iam_member" "vertex_ai_admin_role" {
   role     = "roles/aiplatform.admin"
 }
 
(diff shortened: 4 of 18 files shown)


apis/task.py
```diff
@@ -223,9 +223,9 @@ def run_model(self) -> DAGNode:
       workload_id = xpk.generate_workload_id(self.task_test_config.benchmark_id)
       run_workload = xpk.run_workload(
           task_id="run_workload",
-          project_id=self.task_gcp_config.project_name,
+          cluster_project=self.task_gcp_config.cluster_project,
```
jonb377, line 226: Does GCPConfig.cluster_project get populated if it's omitted on construction, or could this still carry the default `None` value?

configs/example/xpk.Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Dockerfile for xpk examples in xpk_example_dag.py,
```
jonb377, line 1: I'm curious about the workflow on XPK-based runs - how is the docker image built? Is this handled by the infrastructure, or does the user build and tag the image manually?

deployment/cloud_composer_template.tf
```diff
@@ -83,6 +83,22 @@ resource "google_project_iam_member" "vertex_ai_admin_role" {
   role     = "roles/aiplatform.admin"
 }
 
+resource "google_project_iam_member" "artifactr_registry_admin_role" {
```
jonb377, line 86: nit: `artifactr` -> `artifact`

implementations/utils/xpk.py
```diff
@@ -117,15 +117,17 @@ def wait_for_workload_completion(
 
   # Check status of pods
   if not pods.items:
-    RuntimeError(f"No pod is found for workload selector: {pods}.")
-  print(f"pods: {pods}")
+    # This could happen when workload is in the queue (not initialized yet)
+    logging.info(f"No pod is found for workload selector: {pods}.")
+    return False
 
+  print(f"pods: {pods}")
```
jonb377, line 124: Should this print go through `logging`?

jonb377 (commented):
Looks great Ran! I left a couple of questions to help my understanding.

apis/gcp_config.py
```diff
@@ -16,18 +16,23 @@
...
-    dataset_name: The option of dataset for metrics
+    dataset_name: The option of dataset for metrics.
+    cluster_project: The project of a cluster, i.e. a cluster running xpk workloads.
+    cluster_name: The name of a cluster that has provisioned resources.
   """
 
   project_name: str
   zone: str
   dataset_name: metric_config.DatasetOption
+  cluster_project: Optional[str] = None
```
will-cromar, line 37: I wonder if it makes sense to specialize `GCPConfig` for GKE tests

configs/example/xpk.Dockerfile
will-cromar: @jonb377, is there an existing image that your team normally uses with xpk?

When we built test images for all of our tests, it was just one more component that could break and fall out of sync with nightly.

configs/example/xpk.Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN pip install "jax[tpu]" -f https://storage.googleapis.com/jax-releases/libtpu_releases.html
+RUN pip install --no-cache-dir --upgrade clu tensorflow tensorflow-datasets
+RUN git clone https://github.com/google/flax.git /tmp/flax
+RUN pip install --no-cache-dir flax
+RUN curl https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz > /tmp/google-cloud-sdk.tar.gz
+RUN mkdir -p /usr/local/gcloud \
+  && tar -C /usr/local/gcloud -xvf /tmp/google-cloud-sdk.tar.gz \
+  && /usr/local/gcloud/google-cloud-sdk/install.sh
+ENV PATH $PATH:/usr/local/gcloud/google-cloud-sdk/bin
+RUN gcloud components install kubectl
```
will-cromar, line 14: Does the test image itself need to have `kubectl` installed?

configs/example/xpk.Dockerfile
jonb377: I believe the pattern used in MaxText is for developers to build their own baseline dependency image and iterate on top of that, so there's currently no pre-built nightly image. cc @Obliviour to fact check me.

I don't think it would be too much overhead to have MaxText CI pump out nightly images we could use in the tests, and the test scripts are already baked into the MaxText repo.

jonb377: One other thing I just remembered - if we use the same tag across executions, we need to ensure the XPK cluster has the latest version before running (the workloads won't pull from the server if the image exists locally). @Obliviour do you know if this is already handled for the XPK integration here?

apis/task.py
RissyRan: Thanks Jon! Good question. I didn't put a default value for cluster_project. Actually, I think it's better to add composer_project to indicate the project that hosts env info for post_process step (like [this](https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/48/files#diff-82effdf45298e1c77fe8b16705253cd0896789bcf2074caf4726fed50af68226)). I will merge that change into this one.

configs/example/xpk.Dockerfile
RissyRan: `I'm curious about the workflow on XPK-based runs - how is the docker image built? Is this handled by the infrastructure, or does the user build and tag the image manually?`

Thanks Jon for the question! Currently, the example image is manually built. @chandrasekhard2 has a project about docker image with Airflow, and it's in Q1 2024. Looking forward to his proposal/design.

If MaxText builds it's own images. I guess it can be stored in container repo, and pass it directly to xpk command?

configs/example/xpk.Dockerfile
RissyRan: Good question. The example image is used in 2 places:

1) image for [KubernetesPodOperator](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/implementations/utils/xpk.py#L79),
2) image to trigger E2E model test by [xpk](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/implementations/utils/xpk.py#L69).

I think this is needed by the 1st case. But I will double check.

RissyRan: Yes, it's needed for the operator; otherwise, we will get the error below.

Will also make the comment there.

```
b'WARNING: Accessing a Kubernetes Engine cluster requires the kubernetes commandline\nclient [kubectl]. To install, run\n  $ gcloud components install kubectl\n\nFetching cluster endpoint and auth data.\nCRITICAL: ACTION REQUIRED: gke-gcloud-auth-plugin, which is needed for continued use of kubectl, was not found or is not executable. Install gke-gcloud-auth-plugin for use with kubectl by following https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke\nkubeconfig entry generated for mas-v4-8.\n/bin/sh: 1: kubectl: not found\n'
```

configs/example/xpk.Dockerfile
will-cromar: > I don't think it would be too much overhead to have MaxText CI pump out nightly images we could use in the tests, and the test scripts are already baked into the MaxText repo.

That would be helpful. I'd rather rely on an upstream image for testing. If we're building our own image, we're not really testing from the user's perspective.

apis/gcp_config.py
will-cromar: This makes sense to me. We may have to revisit cases where we use the `composer` project to support running a local airflow orchestrator for testing.

will-cromar (approved):

configs/example/xpk.Dockerfile
jonb377: I've asked in the chat - they seem open to the idea of a nightly build. For now, we can just run the standard installation commands on top of a base image.

jonb377 (approved):
LGTM, thanks Ran!

PR #58, opened 2023-12-21
[WIP] Add multislice PyTorch Llama perf tests

jonb377:
# Description

This PR is an exploration to create a workload to run PyTorch/XLA Llama2 perf tests using XPK.

XPK runs on GKE and requires a docker image to run the workloads, which should be up-to-date with the latest dependencies. The proposed flow in this PR is:
- Test authors will add a Dockerfile which prepares the image to be used in their DAGs
- Each DAG invocation will build a docker image using `DockerBuildTask`, which builds an image tagged with the timestamp of the DAG's execution
- The resulting docker image is used in subsequent `TpuXpkTask`s in the DAG

The docker image is built using Cloud Build, which allows for customization using a `cloudbuild.yaml` file to parametrize the build. This PR only builds a Dockerfile directly.

Additionally, at-scale multislice tests are more expensive than single-slice workloads. These tests should run less frequently than single-slice workloads, and they should only be triggered if the smaller-scale single-slice variants are passing. To achieve this, I'm using `ExternalTaskSensor` to watch for success on single-slice tests before allowing multislice to run.

# Tests

Example run of the complete DAG: http://shortn/_KDSU60ZcEr

# Checklist


configs/xlml/pytorch/multipodteam_llama2.py
```diff
@@ -0,0 +1,123 @@
...
+  run_model_cmds = (*hf_setup_commands, train_command)
+
+  job_test_config = test_config.TpuGkeTest(
+    test_config.Tpu(
+      version=tpu_version,
+      cores=tpu_cores,
+    ),
+    test_name=f"pytorch_spmd_llama2_{config_name}_perf",
+    cluster_name=cluster_name,
+    docker_image=vm_resource.DockerImage.PYTORCH_MULTISLICE_BASELINE.value,
```
jonb377, line 103: cc @RissyRan, the image here is being used to launch a container to run `xpk.py workload create` within my test project: http://shortn/_ABDbVB3KIF

I expected this image to be used only for the actual workload on the TPU VMs, so the permissions aren't configured properly to pull it.

RissyRan: Hi Jon, thanks for bring this up!

We should have 2 images. [1st image](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/implementations/utils/xpk.py#L80) is used for pod in env to work around Python3.10 used by xpk, which is not supported in Composer yet. [2nd image](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/implementations/utils/xpk.py#L70) is used for xpk experiments.

Like you mentioned, we could just update the 1st one to python3.10:slim.

PR #80, opened 2024-01-18
Add MaxText parameter sweep configs/dags on GCE and GKE

raymondzouu:
# Description

Add configs/dags for launching parameter job sweeps with MaxText on GCE and GKE.
* `maxtext_sweep_gce_config` and `maxtext_sweep_gke_config` will take in test settings (project_name, tpu_type, num_slices, etc.) and maxtext sweep parameters (per_device_batch_size, remat_policy, ici_tensor_parallelism, etc.), generate all combinations of sweep parameter configurations, and then launch QR jobs or XPK workloads for all configurations
* `maxtext_sweep_gke_example_dag` and `maxtext_sweep_gce_example_dag` are example dags for using these configs

# Tests
* Successfully ran `maxtext_sweep_gce_example_dag` which launched 3 MaxText QR jobs on 1xv4-8 varying the per_device_batch_size
* Successfully ran `maxtext_sweep_gke_example_dag` which launched 3 MaxText GKE/XPK workloads on 1xv4-128 varying the per_device_batch_size
# Checklist

### DIFF
--- a/dags/examples/maxtext_sweep_gce_example_dag.py
+++ b/dags/examples/maxtext_sweep_gce_example_dag.py
@@ -1,0 +1,63 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+An example DAG to launch a sweep of MaxText GCE QueuedResource jobs
+sweeping over a range of per_device_batch_size values for a 1B
+model on 1xv4-8.
+"""
+import datetime
+from airflow import models
+from dags import test_owner
+from dags.vm_resource import TpuVersion, Zone, Project, RuntimeVersion
+from dags.multipod.configs import maxtext_sweep_gce_config
+from dags.multipod.configs import common
+
+
+with models.DAG(
+    dag_id="maxtext_sweep_gce_example_dag",
+    schedule=None,
+    tags=["multipod_team", "maxtext"],
+    start_date=datetime.datetime(2024, 1, 10),
+    catchup=False,
+    concurrency=2,
+) as dag:
+  # MaxText set up and run commands
+  base_set_up_cmds = common.download_maxtext()
+  base_run_model_cmds = [
+      "cd /tmp/maxtext",
+      "bash setup.sh MODE=stable",
+      "python3 MaxText/train.py MaxText/configs/base.yml base_output_directory=gs://maxtext-experiments-multipod/ dataset_path=gs://max-datasets-rogue enable_checkpointing=false global_parameter_scale=1 steps=10",
+  ]
+
+  # Get list of MaxText GCE QueuedResource jobs
+  maxtext_sweep_gce_test = maxtext_sweep_gce_config.get_maxtext_sweep_gce_config(
+      test_owner=test_owner.RAYMOND_Z,
+      project_name=Project.TPU_PROD_ENV_MULTIPOD.value,
+      tpu_zone=Zone.US_CENTRAL2_B.value,
+      time_out_in_min=60,
+      is_tpu_reserved=False,
+      tpu_version=TpuVersion.V4,
+      tpu_cores=8,
+      runtime_version=RuntimeVersion.TPU_UBUNTU2204_BASE.value,
+      num_slices=[1],
+      run_name_prefix="maxtext-1b",
+      base_set_up_cmds=base_set_up_cmds,
+      base_run_model_cmds=base_run_model_cmds,
+      sweep_params={"M_PER_DEVICE_BATCH_SIZE": [1, 2, 4]},
+  )
+
+  # Run jobs
+  for test in maxtext_sweep_gce_test:
+    test.run()
--- a/dags/examples/maxtext_sweep_gke_example_dag.py
+++ b/dags/examples/maxtext_sweep_gke_example_dag.py
@@ -1,0 +1,59 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+An example DAG to launch a sweep of MaxText GKE XPK workloads
+sweeping over a range of per_device_batch_size values for a 16B model
+on 1xv4-128.
+"""
+
+import datetime
+from airflow import models
+from dags import test_owner
+from dags.vm_resource import TpuVersion, Zone, Project, ClusterName, DockerImage
+from dags.multipod.configs import maxtext_sweep_gke_config
+
+
+with models.DAG(
+    dag_id="maxtext_sweep_gke_example_dag",
+    schedule=None,
+    tags=["multipod_team", "maxtext"],
+    start_date=datetime.datetime(2024, 1, 10),
+    catchup=False,
+    concurrency=2,
+) as dag:
+  # MaxText set up and run commands
+  base_run_model_cmds = [
+      "python3 MaxText/train.py MaxText/configs/base.yml base_output_directory=gs://maxtext-experiments-multipod/ dataset_path=gs://max-datasets-rogue enable_checkpointing=false global_parameter_scale=16 steps=10",
+  ]
+
+  # Get list of MaxText GKE XPK jobs
+  maxtext_sweep_gke_test = maxtext_sweep_gke_config.get_maxtext_sweep_gke_config(
+      test_owner=test_owner.RAYMOND_Z,
+      project_name=Project.TPU_PROD_ENV_MULTIPOD.value,
+      cluster_name=ClusterName.V4_128_MULTISLICE_CLUSTER.value,
+      tpu_zone=Zone.US_CENTRAL2_B.value,
+      time_out_in_min=60,
+      tpu_version=TpuVersion.V4,
+      tpu_cores=128,
+      num_slices=[1],
+      docker_image=DockerImage.XPK_MAXTEXT_TEST.value,
+      run_name_prefix="maxtext-16b",
+      base_run_model_cmds=base_run_model_cmds,
+      sweep_params={"M_PER_DEVICE_BATCH_SIZE": [2, 4, 8]},
+  )
+
+  # Run jobs
+  for test in maxtext_sweep_gke_test:
+    test.run()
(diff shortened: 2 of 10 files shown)


dags/multipod/configs/maxtext_sweep_gce_config.py
```diff
@@ -0,0 +1,98 @@
...
+  current_datetime = datetime.datetime.now().strftime("%Y-%m-%d-%H-%M-%S")
+  for idx, config in enumerate(itertools.product(*sweep_params_list)):
+    config_dict = {key: value for (key, value) in config}
+
+    # Remove num_slices as a sweep param after combinations have been generated
+    curr_num_slices = config_dict["NUM_SLICES"]
+    del config_dict["NUM_SLICES"]
+
+    # Add MaxText run_name
+    run_name = f"{run_name_prefix}-{curr_num_slices}x{tpu_version.value}-{tpu_cores}-{current_datetime}-{idx}"
```
tonyjohnchen, line 68: Do we need `{idx}` here? It seems like that `{current_datetime}` already makes the `run_name` unique.

raymondzouu: `idx` is needed here since multiple MaxText jobs can be launched and `current_datetime` will be the same for all of them.

tonyjohnchen (approved):
LGTM

dags/multipod/configs/maxtext_sweep_gce_config.py
tonyjohnchen: Oh i see! It's probably good to not use parameters either since there could be a lot of parameters to sweep.

dags/examples/maxtext_sweep_gke_example_dag.py
```diff
@@ -0,0 +1,58 @@
...
+  maxtext_sweep_gke_test = maxtext_sweep_gke_config.get_maxtext_sweep_gke_config(
+      test_owner=test_owner.RAYMOND_Z,
+      project_name=Project.TPU_PROD_ENV_MULTIPOD.value,
+      cluster_name="v4-bodaborg",
+      tpu_zone=Zone.US_CENTRAL2_B.value,
+      time_out_in_min=60,
+      tpu_version=TpuVersion.V4,
+      tpu_cores=128,
+      num_slices=[1],
+      docker_image=f"gcr.io/{Project.TPU_PROD_ENV_MULTIPOD.value}/raymondzou_runner",
```
jonb377, line 50: @raymondzouu I've been experimenting with having a `docker build` step as part of the DAG to keep the image fresh for each run: http://shortn/_RArxnOOBkp

If you think this would be useful for these tests, we can bundle some of your common setup commands into a Dockerfile in this repo and rebuild the image before each config sweep. Let me know what you think, I can land the docker build step separately from my WIP pytorch tests.

This is really only applicable to the XPK tests though, so not blocking.

dags/examples/maxtext_sweep_gke_example_dag.py
```diff
@@ -0,0 +1,58 @@
...
+      run_name_prefix="maxtext-16b",
+      base_run_model_cmds=base_run_model_cmds,
+      sweep_params={"M_PER_DEVICE_BATCH_SIZE": [2, 4, 8]},
+  )
+
+  # Run jobs
+  for test in maxtext_sweep_gke_test:
```
jonb377, line 57: I don't see any interdependency between the tasks in the DAG right now, so all tasks will be launched at once, is that right? There's some benefit to adding dependencies to slow down the test launch rate, since timeouts would be more likely to happen if capacity is tight and all tests launch at once.

dags/multipod/configs/maxtext_sweep_gce_config.py
```diff
@@ -0,0 +1,98 @@
+# Copyright 2023 Google LLC
```
jonb377, line 1: Do you think we could consolidate the GCE and GKE files? It looks like we could make the parameter sweep logic shared between the two and just change how we construct the `Task` as either `TpuGkeTask` or `TpuGceTask`.

jonb377 (commented):
Nice work Raymond!

dags/examples/maxtext_sweep_gke_example_dag.py
```diff
@@ -0,0 +1,58 @@
...
+  # MaxText set up and run commands
+  base_run_model_cmds = [
+      "python3 MaxText/train.py MaxText/configs/base.yml base_output_directory=gs://maxtext-experiments-multipod/ dataset_path=gs://max-datasets-rogue enable_checkpointing=false global_parameter_scale=16 steps=10",
+  ]
+
+  # Get list of MaxText GKE XPK jobs
+  maxtext_sweep_gke_test = maxtext_sweep_gke_config.get_maxtext_sweep_gke_config(
+      test_owner=test_owner.RAYMOND_Z,
+      project_name=Project.TPU_PROD_ENV_MULTIPOD.value,
+      cluster_name="v4-bodaborg",
```
RissyRan, line 44: Nit: we could reuse this variable: https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/dags/vm_resource.py#L69

dags/examples/maxtext_sweep_gke_example_dag.py
RissyRan, line 50: As this is an example dag, do you think it's better we move docker image to a general path/location (so exclude alias)? The test owner is claimed, so people should be able to find the owner if any issue occurs.

dags/multipod/configs/maxtext_sweep_gce_config.py
RissyRan, line 1: nit: update 2023 to 2024. Same comment for other files.

dags/examples/maxtext_sweep_gce_example_dag.py
```diff
@@ -0,0 +1,62 @@
...
+      time_out_in_min=60,
+      is_tpu_reserved=False,
+      tpu_version=TpuVersion.V4,
+      tpu_cores=8,
+      runtime_version=RuntimeVersion.TPU_UBUNTU2204_BASE.value,
+      num_slices=[1],
+      run_name_prefix="maxtext-1b",
+      base_set_up_cmds=base_set_up_cmds,
+      base_run_model_cmds=base_run_model_cmds,
+      sweep_params={"M_PER_DEVICE_BATCH_SIZE": [1, 2, 4]},
```
RissyRan, line 57: Are you planning to read those params from yml files? Not need to be included in this PR.

dags/examples/maxtext_sweep_gke_example_dag.py
RissyRan: Yes, timeout will happen at the same time. One potential workaround is to add `concurrency` at dag level to restrict how many tasks can run at any moment. However, you may need to adjust this param a little bit as we have some tasks as sensor (i.e. wait_for_request_completion), and this process will up and down. So it may lead to a case that a new test is triggered when sensor task is down.

RissyRan (commented):
Thanks @raymondzouu!

dags/multipod/configs/maxtext_sweep_gce_config.py
raymondzouu: I think we could consolidate the files and parameter sweep logic, but I think we would still need 2 separate functions since there are arguments unique to just GCE (runtime_version, etc.) and GKE (cluster_name, etc.)

dags/examples/maxtext_sweep_gke_example_dag.py
raymondzouu: Done.

dags/multipod/configs/maxtext_sweep_gce_config.py
raymondzouu: Done.

dags/examples/maxtext_sweep_gce_example_dag.py
raymondzouu: I'm not quite sure yet. I think it would be nice to have but not a requirement at the moment. I need to figure out how reading in a yml here would work and I think this way of specifying it in the dag is fine for now.

xlml/utils/xpk.py
```diff
@@ -62,6 +62,9 @@ def run_workload(
       f"gcloud config set compute/zone {zone}",
       "git clone https://github.com/google/xpk.git /tmp/xpk",
       "cd /tmp/xpk",
+      "apt-get update && apt-get install -y google-cloud-sdk",
+      "apt-get install -y kubectl",
```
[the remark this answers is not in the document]
RissyRan: Have you tested it? If so, we could just remove those dependencies from example Dockerfile.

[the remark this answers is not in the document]
jonb377: @raymondzouu we should move as much environment prep and dependency installation into the docker image as possible, but since XPK support is still WIP this approach is fine for now.

dags/examples/maxtext_sweep_gke_example_dag.py
raymondzouu: @jonb377 Thanks for pointing this out! I agree it would be worth it to add some dependencies between the tests to lower the launch rate. I'm thinking we can add some sort of batch launcher where the user can specify the number of tests that can be running at any time and it'll create dependencies to enforce it.

@RissyRan Thanks for the suggestion! I added `concurrency` at the dag level and no longer saw any task timeouts. I'm not sure if I understand what you mentioned about how it may affect the task sensors. I did not see any issues with it in my testing so far.

xlml/utils/xpk.py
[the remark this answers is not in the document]
raymondzouu: I tested this and it worked, thanks!

dags/examples/maxtext_sweep_gke_example_dag.py
raymondzouu: @jonb377 I think a docker build step would be really helpful for nightly XLML tests with XPK since it would need a new nightly build in the image every time it runs. For this case, a benefit of the user supplying there own docker image is that they can have there own maxtext/dependency code changes in the image. Would the new docker build step be able to support this?

jonb377: As I have it now, the docker build runs through cloud build, so it can operate directly on a `Dockerfile` or `cloudbuild.yaml` in the repo. I've only worked with the former use case so far, but using a `cloudbuild.yaml` should be flexible enough to build a docker image from a `Dockerfile` in another repo or run a custom build script.

It sounds like the use case you're targeting here is more for the user to build their own image with local changes to supply to the parameter sweep, is that right? If so, I would say that doesn't sound like a good fit for building the image as part of the DAG, so what you have now is fine. Let me know what you think!

raymondzouu: Yep that is right, for this case I think it is iimportant to be able to supply local changes. Sounds good, I will leave it as is. Thank you for bringing this up!

dags/multipod/configs/maxtext_sweep_gce_config.py
jonb377: Yeah, that makes sense to me. It would be nice to centralize the parameter sweeping logic since there's so much overlap, but if it makes the implementation too clunky it's might not be worth it.

RissyRan (approved):
LGTM!

jonb377 (approved):
Also LGTM, great work Raymond! 👏 None of my comments are blocking, feel free to merge.

dags/examples/maxtext_sweep_gke_example_dag.py
raymondzouu: I will add the batch launcher in a later PR. I think I will need some more time to think about how to implement it.

dags/multipod/configs/maxtext_sweep_gce_config.py
raymondzouu: I will also add this in a later PR. I am planning on adding some more sweep logic and will consolidate it then.

PR #77, opened 2024-01-12
Add Multislice support on GCE, and enable maxtext-nightly test on 1,2,4,8 slices of v4-8

tonyjohnchen:
# Description

Please include a summary of relevant context/issue and your changes.

Add support to multislice and then Add maxtext-nightly test for 1,2,4,8 slices of v4-8

# Tests

Please describe the tests that you ran on Cloud VM to verify changes.

**Instruction and/or command lines to reproduce your tests:** ...

**List links for your tests (use go/shortn-gen for any internal link):** ...
Both 1 slice and 2 slices of v4-8 worked!
https://screenshot.googleplex.com/7wAckVfGZbUsz57

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run one-shot tests and provided workload links above if applicable. 
- [x] I have made or will make corresponding changes to the doc if needed.

PR #82, opened 2024-01-20
Add MaxText tests for Llama2, Mistral, and Gamma

RissyRan:
# Description

Add decode test for Llama2, Mistral, and Gamma
* Reuse the end-to-end bash scripts in MaxText repo
* This PR needs to be done along with another PR in MaxText repo - [PR](https://github.com/google/maxtext/pull/379)

# Tests

Upload to Airflow, and test.

* Successful run for three models on both stable and nightly on Airflow: [link](https://screenshot.googleplex.com/3DBMBFEaMgXyz2A)

# Checklist

### DIFF
--- a/dags/multipod/configs/maxtext_gce_config.py
+++ b/dags/multipod/configs/maxtext_gce_config.py
@@ -73,7 +73,54 @@ def get_maxtext_nightly_config(
       set_up_cmds=set_up_cmds,
       run_model_cmds=run_model_cmds,
       time_out_in_min=time_out_in_min,
-      task_owner=test_owner.Tony_C,
+      task_owner=test_owner.TONY_C,
+      num_slices=num_slices,
+  )
+
+  return task.TpuQueuedResourceTask(
+      task_test_config=job_test_config,
+      task_gcp_config=job_gcp_config,
+  )
+
+
+def get_maxtext_end_to_end_test_config(
+    tpu_version: TpuVersion,
+    tpu_cores: int,
+    tpu_zone: str,
+    time_out_in_min: int,
+    test_name: str,
+    test_script: str,
+    test_mode: common.SetupMode,
+    project_name: str = PROJECT_NAME,
+    runtime_version: str = RUNTIME_IMAGE,
+    network: str = "default",
+    subnetwork: str = "default",
+    is_tpu_reserved: bool = True,
+    num_slices: int = 1,
+) -> task.TpuQueuedResourceTask:
+  job_gcp_config = gcp_config.GCPConfig(
+      project_name=project_name,
+      zone=tpu_zone,
+      dataset_name=metric_config.DatasetOption.XLML_DATASET,
+  )
+
+  set_up_cmds = common.setup_maxtext(test_mode)
+  run_model_cmds = (f"cd /tmp/maxtext && bash end_to_end/{test_script}.sh",)
+
+  job_test_config = test_config.TpuVmTest(
+      test_config.Tpu(
+          version=tpu_version,
+          cores=tpu_cores,
+          runtime_version=runtime_version,
+          reserved=is_tpu_reserved,
+          network=network,
+          subnetwork=subnetwork,
+      ),
+      test_name=test_name,
+      set_up_cmds=set_up_cmds,
+      run_model_cmds=run_model_cmds,
+      time_out_in_min=time_out_in_min,
+      task_owner=test_owner.JON_B,
       num_slices=num_slices,
   )
 
--- a/dags/multipod/maxtext_end_to_end.py
+++ b/dags/multipod/maxtext_end_to_end.py
@@ -1,0 +1,52 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""A DAG to run end-to-end MaxText tests."""
+
+
+import datetime
+from airflow import models
+from dags import composer_env
+from dags.vm_resource import TpuVersion, Zone
+from dags.multipod.configs import maxtext_gce_config
+from dags.multipod.configs.common import SetupMode
+
+
+# Run once a day at 4 am UTC (8 pm PST)
+SCHEDULED_TIME = "0 4 * * *" if composer_env.is_prod_env() else None
+
+
+with models.DAG(
+    dag_id="maxtext_end_to_end",
+    schedule=SCHEDULED_TIME,
+    tags=["multipod_team", "maxtext", "stable", "nightly"],
+    start_date=datetime.datetime(2024, 1, 19),
+    catchup=False,
+) as dag:
+  test_name_prefix = "maxtext"
+  test_models = ["llama2", "mistral", "gamma"]
+  test_modes = [SetupMode.STABLE, SetupMode.NIGHTLY]
+
+  for model in test_models:
+    for mode in test_modes:
+      maxtext_gce_config.get_maxtext_end_to_end_test_config(
+          tpu_version=TpuVersion.V4,
+          tpu_cores=8,
+          tpu_zone=Zone.US_CENTRAL2_B.value,
+          time_out_in_min=60,
+          is_tpu_reserved=False,
+          test_name=f"{test_name_prefix}-{mode.value}-{model}",
+          test_script=f"test_{model}",
+          test_mode=mode,
+      ).run()
--- a/dags/multipod/maxtext_nightly.py
+++ b/dags/multipod/maxtext_nightly.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""A DAG to run all supported ML models with the latest JAX/FLAX version."""
+"""A DAG to run MaxText tests with nightly version."""
 
 import datetime
 from airflow import models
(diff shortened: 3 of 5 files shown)


dags/multipod/configs/maxtext_gce_config.py
```diff
@@ -75,7 +75,82 @@ def get_maxtext_nightly_config(
...
+  set_up_cmds = common.setup_maxtext() + (
+      "pip3 install torch",
+      "pip3 install --no-cache-dir -U crcmod",
+  )
+
+  # download the checkpoint, convert it, and run script
+  current_datetime = datetime.datetime.now().strftime("%Y-%m-%d-%H-%M-%S")
+  parent_path = f"{gcs_bucket.XLML_OUTPUT_DIR}/maxtext/stable/{model_name}"
+  export_vars = (
+      "export M_ENABLE_CHECKPOINTING=true && export M_DATASET_PATH=gs://maxtext-dataset &&"
```
rwitten, line 120: I actually think all of this logic should be in MaxText, not in XLML. The idea is that MaxText holds the "test". What configurations run would be controlled by XLML.

dags/test_owner.py
```diff
@@ -30,5 +30,6 @@
 PEI_Z = "Pei Z."
 
 # MaxText
-Tony_C = "Tony C."
+TONY_C = "Tony C."
```
rwitten, line 33: Should be handles right?

rwitten (changes requested):
Two pieces of feedback:
* I think all of the test logic should live in MaxText. This is easier because when writing code in MaxText it is easier to look for the failing test.
* It looks like this test is written and located in a totally different place from the rest of the MaxText test. Why?

dags/test_owner.py
RissyRan: What did you mean by that? This is the test owner for XL ML tests, and Tony has onboarded some Multipod tests into the system. If you are interested, please take a look at this [PR](https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/77).

dags/multipod/configs/maxtext_gce_config.py
RissyRan: Yes, this could also work. The configuration needed by XL ML is script to setup and run training/inference/whatever. I could just write a helper function to call a bash script from MaxText repo.

dags/test_owner.py
RissyRan: Hi @jonb377, I set you as the owner for those tests at this moment. Please let me know if anyone you think will be a better option.

RissyRan:
> Two pieces of feedback:
> 
> * I think all of the test logic should live in MaxText. This is easier because when writing code in MaxText it is easier to look for the failing test.
> * It looks like this test is written and located in a totally different place from the rest of the MaxText test. Why? (Also, if we're splitting tests to get onto a new system, I want to understand that system!)

Thanks for the feedback! Yes, I understand you want to test logic stay in MaxText, and it should be doable as long as we have a test script the system can directly call. I could update test accordingly next week.

Yes, we have all docs collected on this internal [page](https://g3doc.corp.google.com/cloud/tpu/solutions/g3doc/Tests/index.md?cl=head), please feel free to take a look any part that interested you. And feel free let me know any questions you may have.

dags/multipod/configs/maxtext_gce_config.py
RissyRan: Happy to discuss more details offline! Once we are aligned, I can update the test accordingly. Thank you!

RissyRan: Discussed offline.

dags/multipod/configs/maxtext_gce_config.py
```diff
@@ -75,7 +75,53 @@ def get_maxtext_nightly_config(
...
+    num_slices: int = 1,
+) -> task.TpuQueuedResourceTask:
+  job_gcp_config = gcp_config.GCPConfig(
+      project_name=project_name,
+      zone=tpu_zone,
+      dataset_name=metric_config.DatasetOption.XLML_DATASET,
+  )
+
+  set_up_cmds = common.setup_maxtext()
+  run_model_cmds = (f"bash /tmp/maxtext/end_to_end/{test_script}.sh /tmp/maxtext",)
```
rwitten, line 109: My other comment was just replace this with: `cd /tmp/maxtext/ && bash end_to_end/{test_script}.sh

dags/multipod/maxtext_stable.py
```diff
@@ -0,0 +1,71 @@
...
+with models.DAG(
+    dag_id="maxtext_stable",
+    schedule=SCHEDULED_TIME,
+    tags=["multipod_team", "maxtext", "stable"],
+    start_date=datetime.datetime(2024, 1, 19),
+    catchup=False,
+) as dag:
+  test_name_prefix = "maxtext-stable"
+
+  maxtext_stable_llama2_v4_8 = maxtext_gce_config.get_maxtext_stable_config(
```
rwitten, line 38: We really want these to run on both nightly and stable if possible. Generally, when folks write tests that is how I want them run so we should try to make that super easy.

rwitten (changes requested):

dags/multipod/configs/maxtext_gce_config.py
RissyRan: Yes, good call! Thanks for pointing this out! Updated.

dags/multipod/maxtext_end_to_end.py
```diff
@@ -0,0 +1,112 @@
...
+with models.DAG(
+    dag_id="maxtext_end_to_end",
+    schedule=SCHEDULED_TIME,
+    tags=["multipod_team", "maxtext", "stable", "nightly"],
+    start_date=datetime.datetime(2024, 1, 19),
+    catchup=False,
+) as dag:
+  test_name_prefix = "maxtext"
+
+  maxtext_stable_llama2_v4_8 = maxtext_gce_config.get_maxtext_end_to_end_test_config(
```
rwitten, line 39: This is fine but it seems duplicative -- I think all tests should be automatically set up to run on both stable and nightly.

rwitten (approved):

dags/multipod/maxtext_end_to_end.py
RissyRan: I see. If that is the case, we could have a for loop make it more concise. 

The benefit for this implementation is we can define any run order of tests (if needed), i.e. `test 1-stable >> test 2-stable >> test 3-stable` or `test1-stable >> test1-nightly`.

cc @jonb377

RissyRan: @rwitten, will this for loop change make it more concise? Let me know your thoughts!

dags/multipod/maxtext_end_to_end.py
```diff
@@ -0,0 +1,52 @@
...
+    tags=["multipod_team", "maxtext", "stable", "nightly"],
+    start_date=datetime.datetime(2024, 1, 19),
+    catchup=False,
+) as dag:
+  test_name_prefix = "maxtext"
+  test_models = ["llama2", "mistral", "gamma"]
+  test_modes = [SetupMode.STABLE, SetupMode.NIGHTLY]
+
+  for model in test_models:
+    for mode in test_modes:
```
jonb377, line 42: Could we add a concurrency limit or some dependencies between the tests? I'm always paranoid about capacity failures.

dags/test_owner.py
jonb377: Rafi has a good point - maybe we can follow up on this. It would be nice to have a way to find the test owner's LDAP.

dags/multipod/maxtext_nightly.py
```diff
@@ -28,7 +28,7 @@
 with models.DAG(
     dag_id="maxtext_nightly",
```
jonb377, line 29: How are we logically dividing the DAGs, given the end-to-end tests run using `nightly` as well? Maybe this DAG's ID is a misnomer and it should be `maxtext_nightly_step_time`? cc @tonyjohnchen in case you have any thoughts on this.

dags/multipod/configs/maxtext_gce_config.py
```diff
@@ -75,7 +75,54 @@ def get_maxtext_nightly_config(
...
+    num_slices: int = 1,
+) -> task.TpuQueuedResourceTask:
+  job_gcp_config = gcp_config.GCPConfig(
+      project_name=project_name,
+      zone=tpu_zone,
+      dataset_name=metric_config.DatasetOption.XLML_DATASET,
+  )
+
+  set_up_cmds = common.setup_maxtext(test_mode)
+  run_model_cmds = (f"cd /tmp/maxtext && bash end_to_end/{test_script}.sh",)
```
jonb377, line 110: Just for my understanding - I see that we `cd /tmp/maxtext` in both the `set_up_cmds` and `run_model_cmds`. Do these two execute in different shells?

With @tonyjohnchen's startup-script change, the two currently run in the same shell, so if we expect `run_model_cmds` execute independently of `set_up_cmds` that will need a small adjustment.

jonb377 (commented):
Looks great, thanks Ran!

dags/multipod/maxtext_nightly.py
tonyjohnchen: I used `maxtext_nightly` at first since it should run maxtext with libtpu-nightly every day. I am open to change the name since we will add more tests for maxtext with libtpu-nightly

dags/multipod/configs/maxtext_gce_config.py
RissyRan: Yes, they implement it separately for set_up and run_model.

However in Tony's startup_script implementation, both scripts are merged into one main command, and passed into the startup script. I think it should be fine.

dags/multipod/maxtext_end_to_end.py
RissyRan: Discussed offline. It seems not a major concern for now, as we have ~160 cores quota for v4 TPUs. But we may consider to add concurrency limit in the future when the num of tests increase.

jonb377 (approved):
LGTM

PR #85, opened 2024-01-23
Upsert metric for execution

RissyRan:
# Description

# Tests

# Checklist


xlml/utils/metric.py
```diff
@@ -522,5 +530,8 @@ def process_metrics(
 
   print("Test run rows:", test_run_rows)
 
-  if is_valid_entry():
-    bigquery_metric.insert(test_run_rows)
+  # if is_valid_entry():
```
will-cromar, line 533: Did you mean to remove or uncomment this line?

xlml/utils/bigquery.py
```diff
@@ -110,6 +108,47 @@ def is_valid_metric(self, value: float):
     invalid_values = [math.inf, -math.inf, math.nan]
     return not (value in invalid_values or math.isnan(value))
 
+  def delete(self, row_ids: Iterable[str]) -> None:
+    """Delete records from tables.
+
+    There is a known issue that you cannot delete or update over table
```
will-cromar, line 114: Concretely, what sort of errors would this cause in our case? Duplicate entries?

will-cromar (commented):
Unfortunately, BQ doesn't support upserts directly. I found a couple of answers on the forums and stack overflow recommending using `MERGE` to mimic an upsert. Would this be more consistent than check/delete/insert?  https://www.googlecloudcommunity.com/gc/Data-Analytics/How-to-UPSERT-in-BigQuery/m-p/547777

xlml/utils/metric.py
RissyRan: Thanks Will for taking a look! I am still in progress on this PR, and I think it has automatically sent to your inbox due to code owner file :)

xlml/utils/bigquery.py
RissyRan: Yeah, at high level, in prod stage, I think it should only produce 1 metric record for an execution (i.e. a dag is scheduled on daily basis, and there should only have 1 record from day 2024-1-4, and 1 record from day 2024-1-5 etc. In the current implementation, if we clear the old execution, and a new metric record will be inserted into the BigQuery table. Please see more context in b/318896353.

RissyRan:
> Unfortunately, BQ doesn't support upserts directly. I found a couple of answers on the forums and stack overflow recommending using `MERGE` to mimic an upsert. Would this be more consistent than check/delete/insert? https://www.googlecloudcommunity.com/gc/Data-Analytics/How-to-UPSERT-in-BigQuery/m-p/547777

I see that thread too. But from the BigQuery official tutorial, Merge statement won't work ([link](https://screenshot.googleplex.com/5Fr8wwFNs4Et5jN)). But the Storage Write API works (please see [here](https://cloud.google.com/bigquery/docs/write-api#advantages)). If we want to adapt the method, we need to do following 3 steps: 1) open a stream; 2) send a protocol buffer; 3) close the stream (see [here](https://blog.devgenius.io/bigquery-streaming-in-python-db674a6e4a6a)). That seems a big change to our existing system. I doubt if we should onboard it at this moment. Thoughts?

xlml/utils/metric.py
RissyRan: Thanks! Now it's ready for review.

xlml/utils/metric.py
```diff
@@ -523,4 +531,7 @@ def process_metrics(
   print("Test run rows:", test_run_rows)
 
   if is_valid_entry():
+    # delete records from BigQuery tables for the same Airflow run_id (if applies),
+    # then, insert new records from current run.
+    bigquery_metric.delete(row_ids)
```
zpcore, line 536: QQ, the `row_ids` seems to be different between two airflow run because the `base_id` is different. What are we trying to delete here?

RissyRan: I think I update that logic already, and now it should be the same. Please take a look at here: https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/85/files#diff-74b29a75aab611eb1dff01dac1600bc202aaf83cc6376e182e00676aa6ff06ccR349

xlml/utils/bigquery.py
RissyRan: Discussed offline. Will go through the approach below to avoid data deletion and 90 min buffer:

* Update uuid to be unique for a specific execution on Airflow
* Create a view table for executions with the latest timestamp: so duplicates are still available in original datasets and view table only has 1 record for each exectuion

RissyRan: After a second thought, we could create that view table for job_history as it has timestamp, but we cannot do that for metric_history and metadata_history. 

Schema for metric_history: job_uuid, metric_key, and metric_value
Schema for metadata_history: job_uuid, metadata_key, and metadata_value

PR #86, opened 2024-01-24
Fix SE tests

chandrasekhard2:
# Description

# Tests

# Checklist

### DIFF
--- a/dags/solutions_team/configs/tensorflow/common.py
+++ b/dags/solutions_team/configs/tensorflow/common.py
@@ -54,8 +54,15 @@ FEATURE_TIMEOUT = {
 }
 
 
-def set_up_pjrt_nightly() -> Tuple[str]:
-  """Common set up for PJRT nightly tests."""
+def set_up_se_nightly() -> Tuple[str]:
+  """ Adjust grpc_tpu_worker for SE tests """
+  return (
+      "sudo sed -i 's/TF_DOCKER_URL=.*/TF_DOCKER_URL=gcr.io\/cloud-tpu-v2-images-dev\/grpc_tpu_worker:nightly\"/' /etc/systemd/system/tpu-runtime.service",
+      "sudo systemctl daemon-reload && sudo systemctl restart tpu-runtime",
+  )
+
+def install_tf_nightly() -> Tuple[str]:
+  """Install tf nightly + libtpu."""
   return (
       "pip install tensorflow-text-nightly",
       "sudo gsutil -m cp gs://cloud-tpu-v2-images-dev-artifacts/tensorflow/tf-nightly/latest/*.whl /tmp/ && pip install /tmp/tf*.whl --force",
--- a/dags/solutions_team/configs/tensorflow/solutionsteam_tf_nightly_supported_config.py
+++ b/dags/solutions_team/configs/tensorflow/solutionsteam_tf_nightly_supported_config.py
@@ -61,8 +61,8 @@ def get_tf_keras_config(
       ),
   )
   common_cmds = None
-    if not is_pjrt and is_pod:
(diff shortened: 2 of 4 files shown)


RissyRan (commented):
Thanks Chandra! I feel probably adding `common_cmds` param to address TF specific issue may not be a good idea. To me, `common_cmds` is part of set_up commands, and we could pass them there.

The issue here is, in pod we want to run setup cmds on all workers, while run_model cmds on worker-0 (please correct me if I am wrong). If this is the case, a better way to do this is:

1) split this [all_workers](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/task.py#L60C3-L60C14) into 2: setup_all_workers and run_model_all_workers;
2) Update this [line](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/task.py#L114) to use setup_all_workers
3) Update this [link](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/task.py#L145) to use run_model_all_workers.

I think those could resolve our problem. Please let me know your thoughts!

chandrasekhard2:
> Thanks Chandra! I feel probably adding `common_cmds` param to address TF specific issue may not be a good idea. To me, `common_cmds` is part of set_up commands, and we could pass them there.
> 
> The issue here is, in pod we want to run setup cmds on all workers, while run_model cmds on worker-0 (please correct me if I am wrong). If this is the case, a better way to do this is:
> 
> 1. split this [all_workers](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/task.py#L60C3-L60C14) into 2: setup_all_workers and run_model_all_workers;
> 2. Update this [line](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/task.py#L114) to use setup_all_workers
> 3. Update this [link](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/task.py#L145) to use run_model_all_workers.
> 
> I think those could resolve our problem. Please let me know your thoughts!

Hi Ran, I thought about it but the only concern is that setup_cmds have few git clones and pip installs which we don't need on every worker. changing something on set_up_cmds would also affect pythorch, jax and other tests which need to be reviewed again.

RissyRan:
> > Thanks Chandra! I feel probably adding `common_cmds` param to address TF specific issue may not be a good idea. To me, `common_cmds` is part of set_up commands, and we could pass them there.
> > The issue here is, in pod we want to run setup cmds on all workers, while run_model cmds on worker-0 (please correct me if I am wrong). If this is the case, a better way to do this is:
> > 
> > 1. split this [all_workers](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/task.py#L60C3-L60C14) into 2: setup_all_workers and run_model_all_workers;
> > 2. Update this [line](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/task.py#L114) to use setup_all_workers
> > 3. Update this [link](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/task.py#L145) to use run_model_all_workers.
> > 
> > I think those could resolve our problem. Please let me know your thoughts!
> 
> Hi Ran, I thought about it but the only concern is that setup_cmds have few git clones and pip installs which we don't need on every worker. changing something on set_up_cmds would also affect pythorch, jax and other tests which need to be reviewed again.

Thanks Chandra! I think we could set default value to run on all workers for both setup and run_model cmd. At system level, I'd like to keep the system generality. Happy to discuss your concern offline if that helps.

PR #88, opened 2024-01-25
Fix se tests

chandrasekhard2:
# Description

This PR fixes the Tensorflow multi-host nightly tests.

# Tests

ResNet v4-32.

Run Tensorflow dag

screen/9RZMcHgJRp66ixp

# Checklist

### DIFF
--- a/dags/solutions_team/configs/tensorflow/common.py
+++ b/dags/solutions_team/configs/tensorflow/common.py
@@ -53,12 +53,14 @@ FEATURE_TIMEOUT = {
     TRANSFER_LEARNING: 60,
 }
 
+
 def set_up_se_nightly() -> Tuple[str]:
-  """ Adjust grpc_tpu_worker for SE tests """
+  """Adjust grpc_tpu_worker for SE tests"""
   return (
       "sudo sed -i 's/TF_DOCKER_URL=.*/TF_DOCKER_URL=gcr.io\/cloud-tpu-v2-images-dev\/grpc_tpu_worker:nightly\"/' /etc/systemd/system/tpu-runtime.service",
       "sudo systemctl daemon-reload && sudo systemctl restart tpu-runtime",
   )
+
 
 def install_tf_nightly() -> Tuple[str]:
   """Install tf nightly + libtpu."""
@@ -67,6 +69,7 @@ def install_tf_nightly() -> Tuple[str]:
       "sudo gsutil -m cp gs://cloud-tpu-v2-images-dev-artifacts/tensorflow/tf-nightly/latest/*.whl /tmp/ && pip install /tmp/tf*.whl --force",
       "sudo gsutil -m cp gs://cloud-tpu-v2-images-dev-artifacts/libtpu/latest/libtpu.so /lib/",
   )
+
 
 def set_up_tensorflow_keras() -> Tuple[str]:
   """Common set up for tensorflow Keras tests."""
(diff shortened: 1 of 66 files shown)


scripts/format-codes.sh
```diff
@@ -21,7 +21,7 @@ FOLDERS_TO_FORMAT=("dags" "xlml")
 
 for folder in "${FOLDERS_TO_FORMAT[@]}"
 do
-  pyink "$folder" --pyink-indentation=2 --pyink-use-majority-quotes
+  ~/py39/bin/python3 -m  pyink "$folder" --pyink-indentation=2 --pyink-use-majority-quotes
```
EricLeFort, line 24: Why didn't the old version work, does the PATH not include the installed Python binaries?

Also, we should avoid linking to a specific Python version if possible.

chandrasekhard2: oops, I need to revert this. On my local machine I needed to create a pyenv to install this package.

chandrasekhard2:
Latest successful run - screen/9RZMcHgJRp66ixp .

dags/solutions_team/configs/tensorflow/common.py
```diff
@@ -54,8 +54,17 @@
...
 
 
-def set_up_pjrt_nightly() -> Tuple[str]:
-  """Common set up for PJRT nightly tests."""
+def set_up_se_nightly() -> Tuple[str]:
+  """Adjust grpc_tpu_worker for SE tests"""
+  return (
+      "sudo sed -i 's/TF_DOCKER_URL=.*/TF_DOCKER_URL=gcr.io\/cloud-tpu-v2-images-dev\/grpc_tpu_worker:nightly\"/' /etc/systemd/system/tpu-runtime.service",
+      "sudo systemctl daemon-reload && sudo systemctl restart tpu-runtime",
+      "cat /etc/systemd/system/tpu-runtime.service",
```
RissyRan, line 62: Nit: is this for debugging purpose?

chandrasekhard2: Yes, to explicitly understand which docker image is being used for this test.

RissyRan (approved):

dags/solutions_team/configs/tensorflow/solutionsteam_tf_nightly_supported_config.py
```diff
@@ -142,7 +147,7 @@ def get_tf_resnet_config(
           " PYTHONPATH='.' TF_USE_LEGACY_KERAS=1"
           " python3 official/vision/train.py"
           f" --tpu={tpu_name} --experiment=resnet_imagenet"
-          " --mode=train_and_eval --model_dir=/tmp/output"
+          " --mode=train_and_eval --model_dir=/tmp/"
```
EricLeFort, line 150: Nit: why did we change this? It would be cleaner to sequester the output into a separate folder from the rest of tmp IMO

chandrasekhard2: The plan is to forward the logs to GCS bucket but that change will be in another PR.

EricLeFort: Gotcha, I'm ok with this in the meantime. It's not really a big deal anyway.

PR #90, opened 2024-01-26
Add DockerBuildTask to prepare docker images

jonb377:
# Description

For XPK tests, the docker image should be freshly built with the latest dependencies at the time of test execution. This change adds a DockerBuildTask to build an image using [Cloud Build](https://cloud.google.com/sdk/gcloud/reference/builds/submit).

# Tests

- Integrated with the existing XPK example
- Triggered the DAG for a oneshot: http://shortn/_x9vg8YrnDC
 
# Checklist


RissyRan (commented):
This is great! I noticed the `wait_for_workload_completion` step failed with error. Do you know the reason ([link](https://screenshot.googleplex.com/6xXeUVjdkFeUoot))? I think 4xv4-128 was a known issue (always get stuck and waiting for resource), but not v4-8. Do you happen to know? i.e. some clues from workload link from xpk?

xlml/apis/task.py
```diff
@@ -38,6 +41,72 @@ def run() -> DAGNode:
...
+  task_id: Optional[str] = None
+
+  # Path on the Airflow worker which contains the Airflow GCS bucket contents.
+  GCS_PATH = "/home/airflow/gcs/dags"
+  DEFAULT_TASK_ID = "docker_image_build"
+  IMAGE_NAME_SUBSTITUTION = "_IMAGE_NAME"
+  BUILD_SCRIPT = """
+  set -xe
+  cd {build_path}
+  gcloud builds submit --machine-type=e2-highcpu-32 {build_flags} .
```
RissyRan, line 81: For machine-type, are we able to define the option like [this](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/pipeline/auto-push.cloudbuild.yml#L27)?

dags/examples/xpk_example_dag.py
```diff
@@ -31,14 +32,19 @@
     start_date=datetime.datetime(2023, 11, 29),
     catchup=False,
 ) as dag:
+  docker_image_build = task.DockerBuildTask(
+      build_dir="dags/examples/configs/build",
+      image_name=DockerImage.XPK_JAX_TEST.value,
+      custom_build=True,
```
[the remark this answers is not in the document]
jonb377: Sure, this XPK example is actually better-suited to the direct docker build, since it's just a Dockerfile.

@raymondzouu can I add a docker build step to your parameter sweep example to build MaxText?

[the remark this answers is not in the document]
jonb377: Actually, parameter sweep is probably the one type of XPK test where we don't want a docker build. I'll just add a new DAG.

xlml/apis/task.py
jonb377: We could, but that requires a `cloudbuild.yaml`. This way, the regular docker builds will also use a beefier host for the builds.

jonb377:
> This is great! I noticed the `wait_for_workload_completion` step failed with error. Do you know the reason ([link](https://screenshot.googleplex.com/6xXeUVjdkFeUoot))? I think 4xv4-128 was a known issue (always get stuck and waiting for resource), but not v4-8. Do you happen to know? i.e. some clues from workload link from xpk?

Ah, I suppose it's related to the reformatting in #80, should be an easy fix: http://shortn/_kyXyaD39AF

RissyRan (commented):
Thank you Jon! This feature is great!

Have 2 minor comments:
1) Have you tested on Airflow with the latest change?
2) Do you think it will better if we update `example_dockerfile` and `example_cloudbuild_yaml` to the `docker_build` and `custom_build` that directly maps to your README.md? If so, you may need to update folder names in that file.

others LGTM!

jonb377:
Thanks @RissyRan! I renamed the build folders. Here's a oneshot run: http://shortn/_x9vg8YrnDC

RissyRan (approved):
Thanks Jon! I see 4xv4-128 is still running and I previously also got issues to get it running due to capacity/priority issue. I am going to approve this as single slice and 2 builds workflows work fine.

raymondzouu (approved):
Nice work Jon! LGTM

xlml/apis/task.py
```diff
@@ -38,6 +41,72 @@ def run() -> DAGNode:
...
+  # The last line of output is pushed to XCom as the task output.
+  echo '{image_tag}'
+  """
+
+  def run(self) -> DAGNode:
```
will-cromar, line 86: What's the reason for making this a `Task`? BashOperator is itself a DAG node, so you can string that together with other test `Task`s as needed. Or, you could inherit from the base XPK task and override the `run` and/or `provision` methods

jonb377: We want to build a single Docker image and share it across the DAG, so my intent was to keep it separate from the other run/provision workflows. Do you think `Task` isn't the right abstraction here?

will-cromar: I think of `Task` as an encapsulation of multiple common steps strung together, which spits out a `DAGNode` so it can interoperate with the rest of airflow. Since you have just a single operator here anyway, I don't think it benefits from being wrapped in a `Task`. 

Since you are doing such a simple call to the GCB API, have you looked at the `google-cloud-build` client library? You could wrap this call in a `@task` [^1]. I am obligated to recommend against any form of log parsing from the `gcloud` CLI https://cloud.google.com/python/docs/reference/cloudbuild/latest

[^1]: Either @RissyRan and I probably made a mistake calling `Task`s "tasks" when they're largely made of `@task`s

jonb377: Gotcha, that makes sense, I'll pull it up into Python.

jonb377:
Synced up with @will-cromar - the main issue for xpk currently is reusing the workload's image to run the `xpk` command, hence the need for custom builds. I'll address that first, and if needed we can keep this PR. Closing this out for now.

xlml/apis/task.py
RissyRan: Client library will be a better choice whenever applies. Also, agreed that we called the `Task` with a group of tasks that should be named as task group. Thanks @will-cromar and @jonb377 for this feature!

PR #92, opened 2024-01-27
Migrate legacy test configs and remove submodule

RissyRan:
# Description

Migrate legacy test configs and remove submodule:
* Jax configs
* PyTorch configs
* Template configs

# Tests

* Jax test - [link]()
* PyTorch test - [link]()

# Checklist


RissyRan:
> Thanks! We can definitely prune the files here. I'll follow up to remove anything non-trivial that needs unraveling. Off the top of my head, you can definitely delete the `bisection` folder.

Sure I will delete them!


> Do you think it's worth keeping the run-oneshot script from the other repo under legacy_test? There's not yet an equivalent way to run an isolated oneshot test with airflow.

I feel it may be not needed. Since we don't set up any GKE clusters, we won't trigger it. But, yes, we will find a better way to run one-shot test with Airflow. We could discuss more in next clean-up on this.

PR #93, opened 2024-01-27
enable H100 test and add option to specify Nvidia driver version

zpcore:
# Description

1.  Add option to specify Nvidia driver version for torchbench;
2.  Enable H100, update location to use SSD to create VM;
3.  Fix metric.py issue when calling `provision.create_queued_resource.wait_for_ready_queued_resource`, which doesn't apply for GPU.

# Tests
https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla-torchbench/grid

Let's see if this time we can pass all 7 platforms!

### DIFF
--- a/xlml/utils/gpu.py
+++ b/xlml/utils/gpu.py
@@ -221,6 +221,7 @@ def create_instance(
   return instance_client.get(project=project_id, zone=zone, instance=instance_name)
 
 
+# TODO(piz): Remove wait_for_extended_operation and use dag sensor instead.
 def wait_for_extended_operation(
     operation: ExtendedOperation, verbose_name: str = "operation", timeout: int = 300
 ) -> Any:
@@ -304,7 +305,7 @@ def create_resource(
     The ip address of the GPU VM.
   """
   image = get_image_from_family(project=image_project, family=image_family)
-  disk_type = f"zones/{gcp.zone}/diskTypes/pd-standard"
+  disk_type = f"zones/{gcp.zone}/diskTypes/pd-ssd"
   disks = [disk_from_image(disk_type, 100, True, image.self_link)]
   metadata = create_metadata({
       # "install-nvidia-driver": "True",
@@ -372,6 +373,7 @@ def ssh_host(ip_address: str, cmds: Iterable[str], ssh_keys: ssh.SshKeys) -> Non
   ssh_group.run(cmds)
 
 
+# TODO(piz): Check why sometime TPU instance doesn't get deleted.
 @task_group
 def delete_resource(instance_name: airflow.XComArg, project_id: str, zone: str):
   @task(trigger_rule="all_done")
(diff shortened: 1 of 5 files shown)


zpcore:
Added @tonyjohnchen for `metric.py` update.

xlml/utils/gpu.py
```diff
@@ -372,6 +373,7 @@ def ssh_host(ip_address: str, cmds: Iterable[str], ssh_keys: ssh.SshKeys) -> Non
   ssh_group.run(cmds)
 
 
+# TODO(piz): Check why sometime TPU instance doesn't get deleted.
```
RissyRan, line 376: Nit: GPU? not TPU

xlml/utils/metric.py
```diff
@@ -424,10 +427,16 @@ def get_gce_job_status(
 
   # GCE SSH method
   if not use_startup_script:
-    # check wait status to see if wait_for_ready_queued_resource step is successful
-    wait_task = current_dag.get_task(
-        task_id=f"{benchmark_id}.provision.create_queued_resource.wait_for_ready_queued_resource"
-    )
+    try:
```
RissyRan, line 430: Could we use if statement to check task exists instead of try catch here? And raise an error if both not the case.

xlml/utils/metric.py
```diff
@@ -406,7 +409,7 @@ def get_gke_job_status(benchmark_id: str) -> bigquery.JobStatus:
 
 
 def get_gce_job_status(
-    task_test_config: test_config.TestConfig[test_config.Tpu], use_startup_script: bool
+    task_test_config: test_config.TestConfig[T], use_startup_script: bool
```
RissyRan, line 412: Will you meet issue when you apply `test_config.TestConfig[Accelerator]`?

xlml/utils/metric.py
zpcore: Sorry I didn't quite understand. Isn't the line 430 try, except clause serving the purpose? Do you mean a different way to check if `wait_for_ready_queued_resource` exists or not?

xlml/utils/metric.py
zpcore: Nice catch, I forgot we have `Accelerator` here. Will do a try once the current run is complete.

xlml/utils/metric.py
```diff
@@ -20,7 +20,7 @@
 import hashlib
 import os
 import re
-from typing import Dict, Iterable, List, Optional
+from typing import Dict, Iterable, List, Optional, TypeVar
```
RissyRan, line 23: Nit: we do not need this anymore.

xlml/utils/gpu.py
```diff
@@ -221,6 +221,7 @@ def create_instance(
   return instance_client.get(project=project_id, zone=zone, instance=instance_name)
 
 
+# TODO(piz): Remove wait_for_extended_operation and use dag sensor instead.
```
RissyRan, line 224: Yes, please! So that it is aligned with TPU implementation!

RissyRan (approved):
General LGTM! One nit comment.

tonyjohnchen (commented):
The metric part looks fine to me. Thanks for checking!

will-cromar (approved):

PR #94, opened 2024-01-29
Add llama finetuning test

A9isha:
# Description

Adding a Llama2-7B finetuning test for both stable and nightly MaxText version

# Tests

I have run all the MaxText tests in addition to this new test

Upload to Airflow, and test.

http://shortn/_1pw9SF5HJm

# Checklist

### DIFF
--- a/dags/multipod/maxtext_end_to_end.py
+++ b/dags/multipod/maxtext_end_to_end.py
@@ -35,18 +35,24 @@ with models.DAG(
     catchup=False,
 ) as dag:
   test_name_prefix = "maxtext"
-  test_models = ["llama2", "mistral", "gamma", "gpt3"]
+  test_models = {
+      "llama2": ["test_llama2", "llama_finetuning_test"],
+      "mistral": ["test_mistral"],
+      "gamma": ["test_gamma"],
+      "gpt3": ["test_gpt3"]
+  }
   test_modes = [SetupMode.STABLE, SetupMode.NIGHTLY]
 
-  for model in test_models:
+  for model in test_models.keys():
     for mode in test_modes:
-      maxtext_gce_config.get_maxtext_end_to_end_test_config(
-          tpu_version=TpuVersion.V4,
-          tpu_cores=8,
-          tpu_zone=Zone.US_CENTRAL2_B.value,
-          time_out_in_min=60,
-          is_tpu_reserved=False,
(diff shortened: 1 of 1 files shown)


dags/multipod/maxtext_end_to_end.py
```diff
@@ -35,18 +35,23 @@
     catchup=False,
 ) as dag:
   test_name_prefix = "maxtext"
-  test_models = ["llama2", "mistral", "gamma"]
+  test_models = {
+      "llama2": ["test_llama2", "llama_finetuning_test"],
```
RissyRan, line 39: Thanks Anisha! Do you think it's better to just add this [script](https://github.com/google/maxtext/blob/main/end_to_end/llama_finetuning_test.sh) into [test_llama2](https://github.com/google/maxtext/blob/main/end_to_end/test_llama2.sh)? So no change is needed in this file.

cc @jonb377 as code owner

jonb377: I'd say it's better to keep these tests independent, so the `end_to_end` tests don't require any configuration.

A9isha: Could you please explain the thought a bit more @jonb377 ?

jonb377: Oh I just mean we should keep all the configuration in `maxtext` and run the test using `bash end_to_end/<test_script>` here in XLML. If we consolidate the two scripts, we'd need to pass in some configuration to the test script at runtime.

I like what you have here @A9isha

jonb377 (approved):
LGTM, thanks Anisha!

dags/multipod/maxtext_end_to_end.py
RissyRan: I am happy to approve this PR! I was thinking if it will be better if we consolidate two scripts, as `test_llama2` already has test for training, decode, and others. Wondering what's the purpose to separate fine-tuning to a different script? Or will it be better if we organize it in a way, i.e. separate training or decode from the test_llama2 as well?

I think it's @jonb377's decision.

RissyRan (approved):

dags/multipod/maxtext_end_to_end.py
A9isha: Let us discuss this offline. Let me start a thread

PR #98, opened 2024-01-31
Add maxtext sweep metrics collection

raymondzouu:
# Description

* Add `prepend_tasks` and `append_tasks` attributes to `TpuQueuedResourceTask`, `TpuXpkTask`, `GpuCreateResourceTask` to allow prepending or appending custom tasks to the main job
* Add tasks to generate run name and tensorboard file location for MaxText jobs
* Enable metric processing to retrieve tensorboard file using regex
* Add test metadata (`project`, `accelerator`, etc.) when writing to BigQuery

# Tests

* Ran in local composer environment
* maxtext_sweep_gce_example_dag: http://shortn/_gbZuZQ4rF3
* maxtext_sweep_gke_example_dag: http://shortn/_xS1vM6bOQG

# Checklist

### DIFF
--- a/dags/examples/maxtext_sweep_gce_example_dag.py
+++ b/dags/examples/maxtext_sweep_gce_example_dag.py
@@ -24,7 +24,8 @@ from dags.vm_resource import TpuVersion, Zone, Project, RuntimeVersion
 from dags.multipod.configs import maxtext_sweep_gce_config
 from dags.multipod.configs import common
 
-
+# Set concurrency to number of workers otherwise tasks may time out
+# if there are more concurrent tasks running at a time than number of workers
 with models.DAG(
     dag_id="maxtext_sweep_gce_example_dag",
     schedule=None,
@@ -34,11 +35,12 @@ with models.DAG(
     concurrency=2,
 ) as dag:
   # MaxText set up and run commands
+  base_output_directory = "gs://maxtext-experiments-multipod"
   base_set_up_cmds = common.download_maxtext()
   base_run_model_cmds = [
       "cd /tmp/maxtext",
       "bash setup.sh MODE=stable",
-      "python3 MaxText/train.py MaxText/configs/base.yml base_output_directory=gs://maxtext-experiments-multipod/ dataset_path=gs://max-datasets-rogue enable_checkpointing=false global_parameter_scale=1 steps=10",
+      f"python3 MaxText/train.py MaxText/configs/base.yml base_output_directory={base_output_directory} dataset_path=gs://max-datasets-rogue enable_checkpointing=false global_parameter_scale=1 steps=10",
   ]
 
   # Get list of MaxText GCE QueuedResource jobs
@@ -51,6 +53,7 @@ with models.DAG(
       tpu_version=TpuVersion.V4,
       tpu_cores=8,
       runtime_version=RuntimeVersion.TPU_UBUNTU2204_BASE.value,
+      base_output_directory=base_output_directory,
       num_slices=[1],
       run_name_prefix="maxtext-1b",
       base_set_up_cmds=base_set_up_cmds,
@@ -60,4 +63,4 @@ with models.DAG(
 
   # Run jobs
   for test in maxtext_sweep_gce_test:
-    test.run()
+    test.run_with_run_name_generation()
--- a/dags/examples/maxtext_sweep_gke_example_dag.py
+++ b/dags/examples/maxtext_sweep_gke_example_dag.py
@@ -24,7 +24,8 @@ from dags import test_owner
 from dags.vm_resource import TpuVersion, Zone, Project, ClusterName, DockerImage
 from dags.multipod.configs import maxtext_sweep_gke_config
 
-
+# Set concurrency to number of workers otherwise tasks may time out
+# if there are more concurrent tasks running at a time than number of workers
 with models.DAG(
     dag_id="maxtext_sweep_gke_example_dag",
     schedule=None,
@@ -34,8 +35,9 @@ with models.DAG(
     concurrency=2,
 ) as dag:
   # MaxText set up and run commands
+  base_output_directory = "gs://maxtext-experiments-multipod"
   base_run_model_cmds = [
-      "python3 MaxText/train.py MaxText/configs/base.yml base_output_directory=gs://maxtext-experiments-multipod/ dataset_path=gs://max-datasets-rogue enable_checkpointing=false global_parameter_scale=16 steps=10",
+      f"python3 MaxText/train.py MaxText/configs/base.yml base_output_directory={base_output_directory} dataset_path=gs://max-datasets-rogue enable_checkpointing=false global_parameter_scale=16 steps=10",
   ]
 
   # Get list of MaxText GKE XPK jobs
@@ -45,6 +47,7 @@ with models.DAG(
       cluster_name=ClusterName.V4_128_MULTISLICE_CLUSTER.value,
       tpu_zone=Zone.US_CENTRAL2_B.value,
       time_out_in_min=60,
+      base_output_directory=base_output_directory,
       tpu_version=TpuVersion.V4,
       tpu_cores=128,
       num_slices=[1],
@@ -56,4 +59,4 @@ with models.DAG(
 
   # Run jobs
   for test in maxtext_sweep_gke_test:
-    test.run()
+    test.run_with_run_name_generation()
--- a/dags/multipod/configs/maxtext_sweep_gce_config.py
+++ b/dags/multipod/configs/maxtext_sweep_gce_config.py
@@ -16,7 +16,6 @@
 
 from xlml.apis import gcp_config, metric_config, task, test_config
 from dags.vm_resource import TpuVersion
-import datetime
 import itertools
 from typing import List, Iterable
 
@@ -32,8 +31,10 @@ def get_maxtext_sweep_gce_config(
     run_name_prefix: str,
     project_name: str,
     runtime_version: str,
+    base_output_directory: str,
     base_set_up_cmds: Iterable[str],
     base_run_model_cmds: Iterable[str],
+    dataset_name: metric_config.DatasetOption = metric_config.DatasetOption.BENCHMARK_DATASET,
     is_tpu_reserved: bool = True,
     network: str = "default",
     subnetwork: str = "default",
@@ -41,7 +42,7 @@ def get_maxtext_sweep_gce_config(
   job_gcp_config = gcp_config.GCPConfig(
       project_name=project_name,
       zone=tpu_zone,
-      dataset_name=metric_config.DatasetOption.XLML_DATASET,
+      dataset_name=dataset_name,
       dataset_project=project_name,
       composer_project=project_name,
   )
@@ -56,17 +57,12 @@ def get_maxtext_sweep_gce_config(
 
   # Generate all combinations of sweep param configurations and create a TpuQueuedResourceTask for each one
   qr_task_list = []
-  current_datetime = datetime.datetime.now().strftime("%Y-%m-%d-%H-%M-%S")
   for idx, config in enumerate(itertools.product(*sweep_params_list)):
     config_dict = {key: value for (key, value) in config}
 
     # Remove num_slices as a sweep param after combinations have been generated
     curr_num_slices = config_dict["NUM_SLICES"]
     del config_dict["NUM_SLICES"]
-
-    # Add MaxText run_name
-    run_name = f"{run_name_prefix}-{curr_num_slices}x{tpu_version.value}-{tpu_cores}-{current_datetime}-{idx}"
-    config_dict["M_RUN_NAME"] = run_name
 
     # Export sweep params as env variables for MaxText to read
     run_model_cmds = [f"export {key}={value}" for (key, value) in config_dict.items()]
@@ -89,9 +85,19 @@ def get_maxtext_sweep_gce_config(
         task_owner=test_owner,
         num_slices=curr_num_slices,
     )
+
+    job_metric_config = metric_config.MetricConfig(
+        tensorboard_summary=metric_config.SummaryConfig(
+            file_location=base_output_directory,
+            aggregation_strategy=metric_config.AggregationStrategy.MEDIAN,
+            use_regex_file_location=True,
+        ),
+    )
+
     qr_task = task.TpuQueuedResourceTask(
         task_test_config=job_test_config,
         task_gcp_config=job_gcp_config,
+        task_metric_config=job_metric_config,
     )
     qr_task_list.append(qr_task)
 
--- a/dags/multipod/configs/maxtext_sweep_gke_config.py
+++ b/dags/multipod/configs/maxtext_sweep_gke_config.py
@@ -16,7 +16,6 @@
 
 from xlml.apis import gcp_config, metric_config, task, test_config
 from dags.vm_resource import TpuVersion
-import datetime
 import itertools
 from typing import List, Iterable
 
@@ -33,13 +32,15 @@ def get_maxtext_sweep_gke_config(
     project_name: str,
     cluster_name: str,
     docker_image: str,
+    base_output_directory: str,
     base_run_model_cmds: Iterable[str],
     base_set_up_cmds: Iterable[str] = None,
+    dataset_name: metric_config.DatasetOption = metric_config.DatasetOption.BENCHMARK_DATASET,
 ) -> List[task.TpuXpkTask]:
   job_gcp_config = gcp_config.GCPConfig(
       project_name=project_name,
       zone=tpu_zone,
-      dataset_name=metric_config.DatasetOption.XLML_DATASET,
+      dataset_name=dataset_name,
       dataset_project=project_name,
       composer_project=project_name,
   )
@@ -54,17 +55,12 @@ def get_maxtext_sweep_gke_config(
 
   # Generate all combinations of sweep param configurations and create a TpuXpkTask for each one
   xpk_task_list = []
-  current_datetime = datetime.datetime.now().strftime("%Y-%m-%d-%H-%M-%S")
   for idx, config in enumerate(itertools.product(*sweep_params_list)):
     config_dict = {key: value for (key, value) in config}
 
     # Remove num_slices as a sweep param after combinations have been generated
     curr_num_slices = config_dict["NUM_SLICES"]
     del config_dict["NUM_SLICES"]
-
-    # Add MaxText run_name
-    run_name = f"{run_name_prefix}-{curr_num_slices}x{tpu_version.value}-{tpu_cores}-{current_datetime}-{idx}"
-    config_dict["M_RUN_NAME"] = run_name
 
     # Export sweep params as env variables for MaxText to read
     run_model_cmds = [f"export {key}={value}" for (key, value) in config_dict.items()]
@@ -85,9 +81,19 @@ def get_maxtext_sweep_gke_config(
         cluster_name=cluster_name,
         docker_image=docker_image,
     )
+
+    job_metric_config = metric_config.MetricConfig(
+        tensorboard_summary=metric_config.SummaryConfig(
+            file_location=base_output_directory,
+            aggregation_strategy=metric_config.AggregationStrategy.MEDIAN,
+            use_regex_file_location=True,
+        ),
+    )
+
     xpk_task = task.TpuXpkTask(
         task_test_config=job_test_config,
         task_gcp_config=job_gcp_config,
+        task_metric_config=job_metric_config,
     )
     xpk_task_list.append(xpk_task)
 
--- a/xlml/apis/metric_config.py
+++ b/xlml/apis/metric_config.py
@@ -60,12 +60,15 @@ class SummaryConfig:
     exclude_tag_patterns: The matching patterns of tags that will be excluded.
       No tag is excluded by default. This pattern has higher prioirty to
       include_tag_pattern.
+    use_regex_file_location: Whether to use file_location as a regex to get the
+      file in GCS.
   """
 
   file_location: str
   aggregation_strategy: AggregationStrategy
   include_tag_patterns: Optional[Iterable[str]] = None
   exclude_tag_patterns: Optional[Iterable[str]] = None
+  use_regex_file_location: bool = False
 
 
 @dataclasses.dataclass
--- a/xlml/apis/task.py
+++ b/xlml/apis/task.py
@@ -22,7 +22,7 @@ import airflow
 from airflow.models.taskmixin import DAGNode
 from airflow.utils.task_group import TaskGroup
 from xlml.apis import gcp_config, metric_config, test_config
-from xlml.utils import gpu, metric, ssh, tpu, xpk, startup_script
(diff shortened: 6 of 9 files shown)


dags/examples/maxtext_sweep_gce_example_dag.py
```diff
@@ -24,7 +24,8 @@
 from dags.multipod.configs import maxtext_sweep_gce_config
 from dags.multipod.configs import common
 
-
+# Set concurrency to number of workers otherwise tasks may time out
+# if there are more concurrent tasks running at a time than number of workers
```
jonb377, line 28: Oh interesting, did you hit a limit on the Airflow side?

xlml/utils/metric.py
```diff
@@ -567,5 +659,5 @@ def process_metrics(
 
   print("Test run rows:", test_run_rows)
 
-  if is_valid_entry():
+  if is_valid_entry() or task_metric_config.write_results_to_bigquery:
```
jonb377, line 662: What's an example metric that's not a valid entry but `write_results_to_bigquery=True`?

dags/multipod/configs/maxtext_sweep_gce_config.py
```diff
@@ -56,16 +60,23 @@ def get_maxtext_sweep_gce_config(
...
     del config_dict["NUM_SLICES"]
 
+    # Generate MaxText run name and tensorboard file location
+    with TaskGroup(group_id=f"generate_maxtext_inputs_{idx}") as tg:
+      run_name = maxtext.generate_maxtext_run_name(
+          run_name_prefix, curr_num_slices, tpu_version.value, tpu_cores, idx
+      )
+      tb_file_location = maxtext.generate_maxtext_tb_file_location(
+          run_name, base_output_directory
+      )
```
jonb377, line 77: This is great! I feel like this should be part of the general test infra rather than making it a `prepend` task, since every test will benefit from having a unique GCS bucket path per run. 

@RissyRan How would this fit into the existing infra?

dags/multipod/configs/maxtext_sweep_gce_config.py
```diff
@@ -89,9 +100,21 @@ def get_maxtext_sweep_gce_config(
         task_owner=test_owner,
         num_slices=curr_num_slices,
     )
+
+    job_metric_config = metric_config.MetricConfig(
+        tensorboard_summary=metric_config.SummaryConfig(
+            file_location=tb_file_location,
+            aggregation_strategy=metric_config.AggregationStrategy.MEDIAN,
+            use_regex_file_location=True,
```
jonb377, line 108: Why can't we just consume the `tb_file_location` here? Is it the resulting `tf.events` file having a random name?

xlml/apis/task.py
```diff
@@ -243,11 +262,15 @@ class TpuXpkTask(BaseTask):
...
     task_metric_config: Metric configs to process metrics.
+    prepend_tasks: Custom tasks to run before running main test job
+    append_tasks: Custom tasks to run after running main test job
   """
 
   task_test_config: test_config.TestConfig[test_config.Tpu]
   task_gcp_config: gcp_config.GCPConfig
   task_metric_config: Optional[metric_config.MetricConfig] = None
+  prepend_tasks: DAGNode = None
+  append_tasks: DAGNode = None
```
jonb377, line 273: I'd like to avoid adding these if possible. Using too many one-offs on the infra side can make things harder to maintain long-term. The use case here (i.e. creating random bucket path) to me seems like it should be common to all tests.

cc @RissyRan since this is more infra-side, keen to hear your thoughts.

dags/examples/maxtext_sweep_gke_example_dag.py
```diff
@@ -34,8 +35,10 @@
     concurrency=2,
 ) as dag:
   # MaxText set up and run commands
+  base_output_directory = "gs://maxtext-experiments-multipod"
   base_run_model_cmds = [
-      "python3 MaxText/train.py MaxText/configs/base.yml base_output_directory=gs://maxtext-experiments-multipod/ dataset_path=gs://max-datasets-rogue enable_checkpointing=false global_parameter_scale=16 steps=10",
+      "git checkout gcs_train_config",
```
jonb377, line 40: Is this for testing?

jonb377 (commented):
Nice work Raymond! I like the approach a lot, I think we should consider adding it to the common infra instead of making it specific to the MaxText tests.

dags/examples/maxtext_sweep_gke_example_dag.py
raymondzouu: Oh yes this is for my local testing, I'll remove this when the maxtext changes get merged in.

dags/multipod/configs/maxtext_sweep_gce_config.py
raymondzouu: Right, the `tf.events` file has a random name and we only know the bucket and directory name.

xlml/utils/metric.py
raymondzouu: Currently, writing metrics to BigQuery is only allowed for prod runs in the `cloud-ml-auto-solutions` project. Using `write_results_to_bigquery=True` will allow manual sweep runs in other projects like `tpu-prod-env-multipod` to write metrics to BigQuery.

dags/examples/maxtext_sweep_gce_example_dag.py
RissyRan: OK, I feel this is the behavior I noticed before. When a sensor task (i.e. wait for queued resource ready) is up, it occupied a worker. Once it's down to wait for next check (during 30s or 60s), the worker kicks off another test and send a QR request. 

We fixed this issue by chain the tests (add dependencies between triggered tests). Or we could request more capacity.

dags/multipod/configs/maxtext_sweep_gce_config.py
RissyRan: Yes, I agree. This could benefit other tests.

xlml/apis/metric_config.py
```diff
@@ -94,3 +97,4 @@ class MetricConfig:
   json_lines: Optional[JSONLinesConfig] = None
   tensorboard_summary: Optional[SummaryConfig] = None
   profile: Optional[ProfileConfig] = None
+  write_results_to_bigquery: bool = False
```
RissyRan, line 100: What's the purpose of this flag?

xlml/apis/task.py
```diff
@@ -73,7 +77,22 @@ def run(self) -> DAGNode:
       run_model = self.run_model(queued_resource, ssh_keys)
       post_process = self.post_process()
       clean_up = self.clean_up(queued_resource)
-      provision >> run_model >> post_process >> clean_up
+
```
RissyRan, line 80: Instead of having flags here, could we have `run_with_prepend_tasks` and `run_with_append_tasks`, which is similar to `run()` and `run_with_startup_script`? So you could directly call those, instead of passing flags.

xlml/apis/task.py
RissyRan: I agree with you, Jon. I also suggest to have two separate run tasks, like `run()` or `run_with_startup_script()`.

xlml/utils/xpk.py
```diff
@@ -62,7 +62,7 @@ def run_workload(
       f"gcloud config set compute/zone {zone}",
       "git clone https://github.com/google/xpk.git /tmp/xpk",
       "cd /tmp/xpk",
-      "apt-get update && apt-get install -y google-cloud-sdk",
+      "apt-get install -y google-cloud-sdk",
```
RissyRan, line 65: Wondering the reason to take off `apt-get update`?

xlml/utils/metric.py
RissyRan: Can we move the logic into `is_valid_entry()`? Basically we whitelist the manual sweep runs.

xlml/utils/metric.py
```diff
@@ -332,6 +376,50 @@ def add_airflow_metadata(
   return metadata
 
 
+def add_test_config_metadata(
```
RissyRan, line 379: Can we add unit tests for this func?

xlml/utils/metric.py
```diff
@@ -251,6 +267,34 @@ def process_tensorboard_summary(
   return [metric_history_rows], [metadata_history_rows]
 
 
+def get_gcs_file_location_with_regex(file_location: str):
```
RissyRan, line 270: Can we add unit tests for this func? Thanks!

xlml/utils/metric.py
```diff
@@ -220,7 +226,17 @@ def process_tensorboard_summary(
     a list of MetadataHistoryRow ofr a test run in a test job.
   """
   uuid = generate_row_uuid(base_id, 0)
-  file_location = summary_config.file_location
+
+  if isinstance(summary_config.file_location, airflow.XComArg):
+    file_location = summary_config.file_location.resolve(get_current_context())
```
RissyRan, line 231: Could you explain a little bit for this file_location?

RissyRan (commented):
Thank you!

xlml/apis/metric_config.py
raymondzouu: This flag allows a dag to override the restriction for writing to BigQuery. If this flag is set true then we will allow the run to write to BigQuery.

xlml/utils/metric.py
raymondzouu: I am using a task `generate_maxtext_tb_file_location` to generate the path to the tensorboard file since it includes the run_name. When the task outputs the file path, it is of type XComArg. I found that I needed to call `resolve` in order to get the value of the XComArg.

xlml/utils/metric.py
raymondzouu: Done!

xlml/apis/task.py
RissyRan: I forget to ask the use cases for `prepend_tasks` and `append_tasks`. Could you explain a little bit more?

raymondzouu: The `prepend_tasks` and `append_tasks` are for adding custom tasks to the main test job. This is so that we can add task dependencies on the config level rather than at the dag level. In this use-case, we need to have tasks to generate the `run_name` and the `tb_file_location` and we need to be able to specify it as a dependency to the main test job. I think it is possible to add the tasks on the dag-level however I don't think it is something we should do since the user would be able to specify a `run_name` without using a task.

xlml/utils/xpk.py
raymondzouu: Good catch! I was meaning to add it back

xlml/utils/metric.py
```diff
@@ -250,6 +266,30 @@ def process_tensorboard_summary(
...
+
+  Args:
+    file_location: File location regex in the form of `gs://<your_bucket>/<your_file_path_regex>`.
+
+  Returns:
+    The file location of the first file that fits the given regex.
+  '''
+  storage_client = storage.Client()
+
+  bucket_name = file_location.split("/")[2]
```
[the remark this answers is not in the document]
raymondzouu: Thanks for the suggestion! I changed it to use urlparse however for some values I still needed to use string manipulation.

xlml/utils/metric.py
raymondzouu: Done!

xlml/utils/metric.py
raymondzouu: Done!

xlml/apis/task.py
raymondzouu: @RissyRan @jonb377 I followed Ran's suggestion for creating a new run function `run_with_run_name_generation()`. I made this change just for the GCE flow so far and if it looks good I will change it for the GKE flow as well.

dags/multipod/configs/maxtext_sweep_gke_config.py
```diff
@@ -54,16 +59,22 @@ def get_maxtext_sweep_gke_config(
...
     del config_dict["NUM_SLICES"]
 
+    with TaskGroup(group_id=f"generate_maxtext_run_name_{idx}") as tg:
+      run_name = maxtext.generate_maxtext_run_name(
+          run_name_prefix, curr_num_slices, tpu_version.value, tpu_cores, idx
+      )
+      tb_file_location = maxtext.generate_tb_file_location(
+          run_name, base_output_directory
+      )
+
```
jonb377, line 74: Should this be covered by `run_with_name_generation`?

xlml/utils/maxtext.py
```diff
@@ -0,0 +1,49 @@
...
+import os
+from airflow.decorators import task
+
+
+@task
+def generate_maxtext_run_name(
+    run_name_prefix, curr_num_slices, tpu_version, tpu_cores, idx
+) -> str:
+  current_datetime = datetime.datetime.now().strftime("%Y-%m-%d-%H-%M-%S")
+  return f"{run_name_prefix}-{curr_num_slices}xv{tpu_version}-{tpu_cores}-{current_datetime}-{idx}"
```
jonb377, line 25: It looks like none of the logic here is specific to maxtext, could we rename the file and this method? Maybe something like `xlml/utils/name_format.py`?

dags/examples/maxtext_sweep_gce_example_dag.py
raymondzouu: I found that if `concurrency` was not set and there were more concurrent tasks than workers, then some tasks would end up failing.

xlml/utils/metric.py
```diff
@@ -251,6 +266,35 @@ def process_tensorboard_summary(
...
+  prefix = "/".join(file_path.split("/")[:-1])
+
+  all_blobs_names = [
+      b.name for b in storage_client.list_blobs(bucket_name, prefix=prefix)
+  ]
+  if not all_blobs_names:
+    logging.error(f"No files found with given regex {file_location}")
+    raise FileNotFoundError
+
+  return f"gs://{bucket_name}/{next(filter(file_path_regex.match, all_blobs_names))}"
```
jonb377, line 295: Could we raise a different error if nothing matches? Right now the error will be `StopIteration`. Maybe:

```python
try:
  return f"gs://{bucket_name}/{next(filter(file_path_regex.match, all_blobs_names))}"
except:
  raise AirflowFailException(f'No objects matched supplied regex: {file_location}')
```

xlml/utils/metric_test.py
```diff
@@ -315,27 +316,116 @@ def test_add_airflow_metadata(self):
...
+        composer_project="test_project",
+    )
+
+    actual_value = metric.add_test_config_metadata(
+        base_id,
+        task_test_config,
+        task_gcp_config,
+        raw_meta,
+    )
+    print("actual_value", actual_value)
```
jonb377, line 362: Do we need these prints?

xlml/utils/metric.py
```diff
@@ -332,6 +376,50 @@ def add_airflow_metadata(
...
+        )
+    )
+    test_config_meta.append(
+        bigquery.MetadataHistoryRow(
+            job_uuid=uuid,
+            metadata_key="project",
+            metadata_value=task_gcp_config.project_name,
+        )
+    )
+    if hasattr(task_test_config, "num_slices"):
```
RissyRan, line 403: Will you meet error without this? I see `TestConfig[Tpu]` is defined here

xlml/apis/task.py
```diff
@@ -77,6 +77,39 @@ def run(self) -> DAGNode:
...
+      run_name = maxtext.generate_run_name(self.task_test_config.benchmark_id)
+      tb_file_location = maxtext.generate_tb_file_location(
+          run_name, self.task_metric_config.tensorboard_summary.file_location
+      )
+
+      # Set run_name in run_model_cmds
+      new_run_model_cmds = [f"export M_RUN_NAME={run_name}"]
+      for cmd in self.task_test_config.run_model_cmds:
+        new_run_model_cmds.append(cmd)
+      self.task_test_config.run_model_cmds = new_run_model_cmds
```
jonb377, line 99: Ah, this part does seem MaxText-specific... Is there any way around modifying the run_model_cmds to communicate the run name to the workload? cc @RissyRan for her AirFlow expertise, maybe set a variable here and consume it in the workload?

jonb377 (approved):
LGTM on the multipod side, a couple of comments on generality but nothing blocking. Great work Raymond!

RissyRan (commented):
Thanks @raymondzouu

xlml/apis/task.py
raymondzouu: Sounds good, I created a new run task as suggested.

dags/multipod/configs/maxtext_sweep_gke_config.py
raymondzouu: Yes good catch, I changed it to use the new `run_with_name_generation`

xlml/utils/maxtext.py
raymondzouu: Done!

xlml/utils/metric_test.py
raymondzouu: I followed the same pattern as the existing tests https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/5e0d91f69f6a68640d3d8be6ffa486e961d3fd7b/xlml/utils/metric_test.py#L285-L293. I think it would be good to have the prints in case the test fails we could see the difference.

xlml/utils/metric.py
```diff
@@ -332,6 +376,50 @@ def add_airflow_metadata(
...
+          bigquery.MetadataHistoryRow(
+              job_uuid=uuid,
+              metadata_key="num_slices",
+              metadata_value=task_test_config.num_slices,
+          )
+      )
+      test_config_meta.append(
+          bigquery.MetadataHistoryRow(
+              job_uuid=uuid,
+              metadata_key="topology",
```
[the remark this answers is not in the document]
raymondzouu: Oh good catch, I changed it to `multislice_topology`. What do you think?

xlml/utils/metric.py
```diff
@@ -251,6 +266,35 @@ def process_tensorboard_summary(
   return [metric_history_rows], [metadata_history_rows]
 
 
+def get_gcs_file_location_with_regex(file_location: str):
```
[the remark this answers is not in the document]
raymondzouu: Done!

xlml/apis/task.py
```diff
@@ -77,6 +77,39 @@ def run(self) -> DAGNode:
 
     return group
 
+  def run_with_run_name_generation(self) -> DAGNode:
+    """Run a test job.
+
+    Returns:
+      A task group with the following tasks chained: provision, run_model,
```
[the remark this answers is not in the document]
raymondzouu: Done!

xlml/utils/metric.py
raymondzouu: I changed the parameter to be `TestConfig[test_config.Accelerator]` since this function would also be called for GPU tests. I think we need the if since there is no `num_slices` in `GpuVmTest` https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/5e0d91f69f6a68640d3d8be6ffa486e961d3fd7b/xlml/apis/test_config.py#L188

xlml/apis/task.py
raymondzouu: Yeah I agree this part is MaxText specific. I think the only way I could think of to keep the generality here was to have the `prepend_tasks` where we could contain the MaxText specific tasks/operations and then prepend it to the main job. I would definitely be open to other ideas or workarounds.

xlml/utils/metric.py
```diff
@@ -355,11 +443,16 @@ def generate_process_id() -> str:
   return str(uuid.uuid4())
 
 
-def is_valid_entry() -> bool:
+def is_valid_entry(task_metric_config: metric_config.MetricConfig) -> bool:
```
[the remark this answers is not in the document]
raymondzouu: Oh I see, no worries! Do you mean to remove the lines mentioned and only allow runs in the prod env `ml-automation-solutions`? How about collecting metrics in individual development environments?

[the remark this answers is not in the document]
RissyRan: If you remove those lines (https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/utils/metric.py#L368-L373), we enabled metric collections for manual runs in prod. So basically all metrics from runs in prod stage will be sent to datasets. I think this is also what you want right? still no metrics from non-prod runs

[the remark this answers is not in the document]
raymondzouu: I would also want to collect metrics in development composer environment. For example if I want to run a sweep of jobs in my own environment and get metrics for it

xlml/apis/task.py
RissyRan: Yeah, I think I get your point Jon. You may want to add a global variable [here](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/test_config.py#L184) to add this export directly. Currently, I did not see other use cases to make changes to run_model_cmds besides maxtext runs. As we already have a `run_with_run_name_generation` helper here to handle the case. I am fine with current implementation :)

xlml/utils/metric.py
RissyRan: Yeah, I like this `AirflowFailException`.

xlml/utils/metric.py
RissyRan: SG! Thanks!

xlml/utils/metric.py
raymondzouu: I did a run of the `flax_bert_mrpc_conv-v4-32` http://shortn/_I2nUHjE4RC in my own development environment in `tpu-prod-env-multipod`. It was able to successfully get the tensorboard file and parse it. There was just an unrelated failure since I was using a different project but it was not because of the new change.

xlml/utils/metric.py
raymondzouu: Thanks for the suggestion, added it!

xlml/utils/metric.py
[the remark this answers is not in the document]
RissyRan: I understand that. I think the right way for this is to have a separate dev dataset setup for dev composer env. We shouldn't pollute dev data into prod stage.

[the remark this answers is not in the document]
raymondzouu: That sgtm, I changed it so that all runs in the prod env (including manual runs) will write to bigquery and then any env can write to a separate dev dataset `benchmark_dataset_dev`. Does this look good?

xlml/apis/task.py
raymondzouu: Sounds good. I am fine with the current implementation as well, and we can modify it later if issues arise.

xlml/utils/metric.py
RissyRan: Sounds good. Thank you!

xlml/utils/metric.py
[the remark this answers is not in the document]
raymondzouu: I removed the changes to `is_valid_entry()`. Will follow the changes from https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/112 as discussed offline.

RissyRan (approved):
Thank you, Raymond!

PR #104, opened 2024-02-05
add compute instance admin role

zpcore:
Add compute.instanceAdmin.v1 role to PROD and DEV environment. The role is for creating VM instance through gcp v1 api.

### DIFF
--- a/deployment/cloud_composer_template.tf
+++ b/deployment/cloud_composer_template.tf
@@ -99,6 +99,14 @@ resource "google_project_iam_member" "monitoring_viewer_role" {
   role     = "roles/monitoring.viewer"
 }
 
+resource "google_project_iam_member" "compute_instance_admin_role" {
+  provider = google-beta
+  for_each = local.environment_config_dict
+  project  = var.project_config.project_name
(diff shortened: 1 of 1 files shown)


will-cromar (approved):
Is this permission for creating/deleting instances from the composer service account?

deployment/cloud_composer_template.tf
```diff
@@ -99,6 +99,14 @@ resource "google_project_iam_member" "monitoring_viewer_role" {
   role     = "roles/monitoring.viewer"
 }
 
+resource "google_project_iam_member" "compute_instance_admin_role" {
```
RissyRan, line 102: Could you help have a quick try to have this role in default service account only? And see if this permission is needed here. Sorry I should mention it earlier.

zpcore:
> Is this permission for creating/deleting instances from the composer service account?

Yes. For example, using the API here: https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/693ee3f03187b7fd2c8d2753245a2072c720c0b0/xlml/utils/gpu.py#L149.

RissyRan (approved):

PR #110, opened 2024-02-12
Add instructions to run a local Airflow server

will-cromar:
# Description

See README diff

# Tests

Tested instructions locally with a WIP DAG.

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,16 +20,6 @@ Thank you for your interest in contributing to this project!
 
 Please review the [contribution guidelines](docs/contributing.md), and note that all contributions must adhere to the [code of conduct](docs/code-of-conduct.md).
 
-### Testing changes locally
-
-To run a standalone Airflow server locally without uploading your changes, run the following from your copy of this repository:
-
-```
-AIRFLOW_HOME=$PWD PYTHONPATH=. airflow standalone
-```
-
-This functionality is extremely experimental, and not all DAGs are expected to work with a local standalone server. Only the Airflow server runs locally. Tests will still run in the project defined in each DAG, so use this option with caution.
-
 ## License
 
 [Apache License 2.0](LICENSE)
--- a/README.md
+++ b/README.md
@@ -20,6 +20,16 @@ Thank you for your interest in contributing to this project!
 
 Please review the [contribution guidelines](docs/contributing.md), and note that all contributions must adhere to the [code of conduct](docs/code-of-conduct.md).
 
+### Testing changes locally
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -20,6 +20,16 @@ Thank you for your interest in contributing to this project!
 
 Please review the [contribution guidelines](docs/contributing.md), and note that all contributions must adhere to the [code of conduct](docs/code-of-conduct.md).
 
+### Testing changes locally
```
RissyRan, line 23: Do you want to move this into this [contributing process step](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/03188791e1548defe894f43aadb81f18df7120d4/docs/contributing.md)? So we have contributing guide together?

RissyRan (commented):
Thanks Will! How will this feature look like? 

I ran this standalone before, and recall I need to do some setup. Is this the same for you? Could you provide one simple example for the run? I think this will be helpful for users as well.

README.md
RissyRan: Or it's actually better if we move that all content from contributing page to the main page [README.md](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/README.md)?

will-cromar: Good idea. I'll move this to the contributing doc.

will-cromar:
> Thanks Will! How will this feature look like?
> 
> I ran this standalone before, and recall I need to do some setup. Is this the same for you? Could you provide one simple example for the run? I think this will be helpful for users as well.

I actually didn't have to do any other setup besides authenticating with GCP `gcloud auth login --update-adc` . `airflow standalone` does the one-time setup and prints a link to a local server. I didn't try all of the DAGs, so I don't expect them all to work. Some might be coupled to the composer environment, or people might not have the right permissions (e.g. to directly create QR in the project).

I'm contributing this as a quick hack that may or may not work, but we should expand it over time to work with more task types.

docs/contributing.md
```diff
@@ -47,4 +47,15 @@ Run `bash scripts/format-codes.sh` to format your codes.
 
 All submissions, including submissions by project members, require review. We
 use [GitHub pull requests](https://docs.github.com/articles/about-pull-requests)
-for this purpose.
\ No newline at end of file
+for this purpose.
+
+### Testing changes locally
```
RissyRan, line 52: Nit: `Testing changes locally` to `Testing Changes Locally` to match the format.

RissyRan (approved):
Approved with a nit.

PR #111, opened 2024-02-12
Multi-host GPU tests on GKE

will-cromar:
# Description

- Implement multi-host GPU tests via GKE. GKE helps with resource management/coordination and provides a service discovery mechanism that we need for `torchrun`.
- Uses legacy test templates, adapting @vanbasten23's PR in the old repository (https://github.com/GoogleCloudPlatform/ml-testing-accelerators/pull/1011) to the new system.
- New task type currently requires the `JSonnetGpuTest`. I intend to clean this up and generalize the API.
- Factor out GKE authentication to `gke.get_authenticated_client`

# Tests

Tested locally. Will post screenshots.

# Checklist


will-cromar:
There's still plenty of cleanup to do here, but the basic idea works!

Screenshot:

<details>

image

</details>

dags/legacy_test/tests/pytorch/nightly/common.libsonnet
```diff
@@ -120,7 +120,36 @@ local volumes = import 'templates/volumes.libsonnet';
...
+        # Run whatever is in `command` here
+        "${@:0}"
+      |||,
+    ],
+    command: [
+      'torchrun',
+      '--nnodes=%d' % config.accelerator.num_hosts,
+      '--node_rank=$(JOB_COMPLETION_INDEX)',
+      '--nproc_per_node=%d' % config.accelerator.count,
+      '--rdzv_endpoint=$(JOB_NAME)-0.headless-svc:12355',
```
vanbasten23, line 151: This is very cool. Thanks Will!

dags/legacy_test/tests/pytorch/nightly/common.libsonnet
```diff
@@ -120,7 +120,36 @@ local volumes = import 'templates/volumes.libsonnet';
...
+    ] + super.command[1:],
```
vanbasten23, line 152: I wonder if there is a development guide for me to read first about how this whole thing works before I can review the PR.

will-cromar: This PR is going to be rough. I'm basically piling one last hack on top of our template system, and then we need to start simplifying and flattening the test configs to make them more approachable.

The big idea here is, we build the test config mostly as normal in JSonnet. We can add a few more tests after this PR just by editing the templates and registering the test name in the dag file. But instead of generating the Kuberenetes manifest within the JSonnet template system, we create it in Python. The airflow orchestrator will then apply the config, watch the status, and stream the logs from GKE.

will-cromar:
Tested again after refactoring:

<details>

image

</details>

xlml/apis/task.py
```diff
@@ -503,3 +504,118 @@ def clean_up(self, resource: airflow.XComArg, project_id: str, zone: str) -> DAG
...
+        "kind": "Job",
+        "metadata": {
+            "generateName": f"{self.task_test_config.benchmark_id}-",
+            "labels": {
+                "accelerator": self.task_test_config.accelerator.name,
+                "benchmarkId": self.task_test_config.benchmark_id,
+            },
+        },
+        "spec": {
+            # "activeDeadlineSeconds": 10800,
```
vanbasten23, line 556: Need to uncomment this?

dags/legacy_test/tests/pytorch/nightly/common.libsonnet
```diff
@@ -120,7 +120,36 @@ local volumes = import 'templates/volumes.libsonnet';
...
+      |||
+        export PATH=/usr/local/nvidia/bin${PATH:+:${PATH}}
+        export LD_LIBRARY_PATH=/usr/local/nvidia/lib64:/usr/local/nvidia/lib${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}
+
+        nvidia-smi
+        pip3 uninstall -y torch torchvision
+        pip3 install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cpu
+
+        mkdir pytorch
+        wget https://github.com/pytorch/xla/archive/refs/heads/master.tar.gz -O - | tar xzf -
```
vanbasten23, line 137: Out of curiosity, why don't we do `git clone ` as before?

dags/legacy_test/tests/pytorch/nightly/common.libsonnet
```diff
@@ -120,7 +120,36 @@ local volumes = import 'templates/volumes.libsonnet';
...
+
+        # Run whatever is in `command` here
+        "${@:0}"
+      |||,
+    ],
+    command: [
+      'torchrun',
+      '--nnodes=%d' % config.accelerator.num_hosts,
+      '--node_rank=$(JOB_COMPLETION_INDEX)',
+      '--nproc_per_node=%d' % config.accelerator.count,
```
vanbasten23, line 150: Can this be a parameter because for SPMD this needs to be 1?

xlml/apis/task.py
```diff
@@ -503,3 +504,118 @@ def clean_up(self, resource: airflow.XComArg, project_id: str, zone: str) -> DAG
...
+            "backoffLimit": 0,
+            "completionMode": "Indexed",
+            "completions": self.task_test_config.num_hosts,
+            "parallelism": self.task_test_config.num_hosts,
+            "template": {
+                "metadata": {
+                    # Matches `headless-svc` in GKE cluster. See deployments directory.
```
vanbasten23, line 563: I wonder where the code creates the service: headless-svc.

vanbasten23:
Looking great! Left some minor comment. Thanks!

.github/requirements.txt
```diff
@@ -7,4 +7,5 @@ google-cloud-container
 google-cloud-tpu>=1.16.0
 jsonlines
 tensorflow-cpu
-apache-airflow-providers-cncf-kubernetes
\ No newline at end of file
+apache-airflow-providers-cncf-kubernetes
+kubernetes
```
RissyRan, line 11: Is this needed for terraform? If not, could we add something like this? https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/deployment/cloud_composer_template.tf#L143

dags/legacy_test/tests/pytorch/nightly/common.libsonnet
RissyRan: Trying to learn more. What's this `headless`?

xlml/apis/task.py
```diff
@@ -503,3 +504,118 @@ def clean_up(self, resource: airflow.XComArg, project_id: str, zone: str) -> DAG
...
+  Attributes:
+    image_project: the project that an image belongs to.
+    image_family: the family group that an image belongs to.
+  """
+
+  task_test_config: test_config.JSonnetGpuTest
+  task_gcp_config: gcp_config.GCPConfig
+  cluster_name: str
+  job_create_timeout: datetime.timedelta = datetime.timedelta(minutes=10)
+  # TODO: metrics
```
RissyRan, line 522: Will you be the task owner of this? Basically add post_process right?

xlml/apis/task.py
```diff
@@ -503,3 +504,118 @@ def clean_up(self, resource: airflow.XComArg, project_id: str, zone: str) -> DAG
...
+  # task_metric_config: Optional[metric_config.MetricConfig] = None
+
+  def run(self) -> DAGNode:
+    """Run a test job.
+
+    Returns:
+      A task group with the following tasks chained: provision, run_model,
```
RissyRan, line 529: I don't think we will have provision and clean_up any more right?

xlml/apis/task.py
```diff
@@ -503,3 +504,118 @@ def clean_up(self, resource: airflow.XComArg, project_id: str, zone: str) -> DAG
...
+      post_process, clean_up.
+    """
+    # piz: We skip the queued resource for GPU for now since there is no queued
+    # resource command for GPU.
+    with TaskGroup(
+        group_id=self.task_test_config.benchmark_id, prefix_group_id=True
+    ) as group:
+      job_body = self._get_job_manifest()
+      gke.run_job(
```
RissyRan, line 538: I understand when more options/workflows come, we will diverge the name a little bit. But I think it will be good if we keep the name aligned as much as possible, i.e. still use run_model and post_process if applies. (

This is one [example](https://screenshot.googleplex.com/BtQsZgw4gsXR3Pz) for GKE on TPU, I am thinking we could rename something below. 

In this case:

```
run_model
* deploy_workload
* stream_logs
post_process
```

In xpk case (I can do the change):

```
run_model
* generate_workload_id
* deploy_workload
* wait_for_workload_completion
post_process
```

I like `workload` better as job can be a end-to-end test job in general.

xlml/apis/task.py
```diff
@@ -503,3 +504,118 @@ def clean_up(self, resource: airflow.XComArg, project_id: str, zone: str) -> DAG
...
+          job_body, self.task_gcp_config, self.cluster_name, self.job_create_timeout
+      )
+
+    return group
+
+  def _get_job_manifest(self):
+    return {
```
RissyRan, line 545: Can we move those to a config/manifest file and read/import here?

xlml/utils/gke.py
```diff
@@ -0,0 +1,132 @@
...
+  @task.sensor(
+      poke_interval=60, timeout=job_create_timeout.total_seconds(), mode='reschedule'
+  )
+  def stream_logs(name: str):
+    client = get_authenticated_client(gcp.project_name, gcp.zone, cluster_name)
+
+    batch_api = kubernetes.client.BatchV1Api(client)
+    job = batch_api.read_namespaced_job(namespace='default', name=name)
+
+    # TODO: Handle other conditions (e.g. unschedulablility)
```
RissyRan, line 66: Nit: owner if we know? :)

RissyRan (commented):
Thanks Will for this feature! This is great!

I have seen a successful run already, just check you also have a failed run that being captured :)

dags/legacy_test/tests/pytorch/nightly/common.libsonnet
will-cromar: This ends up being somewhat faster to run since we don't have to download the whole git history. We should actually be putting this on our release images IMO.

dags/legacy_test/tests/pytorch/nightly/common.libsonnet
will-cromar: This is a great point that I haven't thought of. Is it okay if I handle this case in a follow up? There are other aspects I want to clean up (particularly how we handle the setup script), but this PR is already large and I don't want to include too much template refactoring.

xlml/apis/task.py
will-cromar: Currently with the cluster definition in `clusters.tf`. I didn't point to a specific file since I expect that code to move over time, e.g. if we pull some of the cluster config into a terraform module.

.github/requirements.txt
will-cromar: Good point. This package is preinstalled, so I added a commented dep.

dags/legacy_test/tests/pytorch/nightly/common.libsonnet
vanbasten23: headless-svc is the name of the kubernets service.

xlml/apis/task.py
will-cromar: This resnet test itself doesn't actually write metrics correctly (see the  `--nometrics_debug` flag in the test configs), so I don't actually have a good test case to make sure this works. Otherwise adding this should be straightforward, so whoever adds a new multihost GPU test that writes metrics can enable this.

dags/legacy_test/tests/pytorch/nightly/common.libsonnet
will-cromar: "Headless" is a concept from kubernetes: https://kubernetes.io/docs/concepts/services-networking/service/#headless-services

It gives all pods that match the selector a subdomain under its name. Since the service name is `headless-svc`, matching pods (ie those that define the right label) with names like `abc-123` will be addressable as `abc-123.headless-svc`. This is relevant to multihost GPU jobs, since Kubernetes gives them predictable hostnames like `$JOB_NAME-0`, `$JOB_NAME-1`, etc. Attaching them to a headless service makes them accessible by a predictable DNS name.

xlml/apis/task.py
will-cromar: Yeah, you're right. Took a pass to clean up all of the docstrings.

xlml/apis/task.py
will-cromar: Renaming the inner tasks SGTM. I think the top-level Python name `gke.run_job` makes sense here because it is taking a Kubernetes `Job` here. "Job" ends up being overloaded in general, but I this actually means `Job` in a specific sense, which is distinct from other types of workloads you can run on GKE.

will-cromar: IIRC it is possible to overload the name of a decorated task group, so we could keep the Python API of `gke.run_job` and generate a task group named `run_model`.

xlml/apis/task.py
will-cromar: Ironically, since this is templated based parameters of the job config, the best place to move this as  static config would be JSonnet :laughing: 

I'd like to keep as much "logic" in Python compared to JSonnet moving forward, and this makes the most sense as a private helper since it depends only on the attributes of this class. But `task.py` is getting untenably long and hard to read, and this is a particularly long method because Kubernetes manifests are verbose. I propose two options:

1. Move each task type (or group of similar task types) to new files like `task.gke` or `task.gpu`.
2. Move each task type to the place where their implementations are defined (e.g. this becomes `gke.GpuGkeTask`.

xlml/utils/gke.py
will-cromar: This is probably going to be me

will-cromar:
> Thanks Will for this feature! This is great!
> 
> I have seen a successful run already, just check you also have a failed run that being captured :)

Yeah, I tested that failing exit codes are propagated by interrupting one  of the containers during a test run.

vanbasten23 (approved):
LGTM. Thanks!

dags/legacy_test/tests/pytorch/nightly/common.libsonnet
RissyRan: Thank you @will-cromar for the explanation!

xlml/apis/task.py
RissyRan: Oh...I did not mean the benchmarking metrics. I was saying the default metric entry into job_history. Basically integrate with post_process step so that in the dashboard (at least for system monitoring purpose), i.e. we know this tests succeeded or failed, and how many tests are running in our repo, etc.

xlml/apis/task.py
RissyRan: Sounds good to me with the task group named run_model. Will we have structure like this on UI?

```
run_model 
* deploy_workload 
* stream_logs 
post_process
```

xlml/apis/task.py
RissyRan: Yes, this is really long now. I may prefer the 1st option, so we will have:

based on accelerators:
```
task_gpu.py
task_tpu.py
``` 

Or based on platforms:

```
task_gke.py
task_gce.py
```

This will be a big refactor, can you add a TODO with my name.

RissyRan (commented):
Thanks Will! Overall looks good!

I have one high level question that we should integrate with post_process step to emit default metrics? Or I missed that in the latest test.

xlml/apis/task.py
will-cromar: Got it. I'll put myself as a task owner for job history metrics.

xlml/apis/task.py
will-cromar: Right. Added the override to name the task group `run_model`

RissyRan (approved):

PR #118, opened 2024-02-15
Add flags to setup and run scripts

RissyRan:
# Description

Add flags to setup and run scripts to avoid hidden failure (b/322868084).

# Tests

Upload to Airflow and test.

Successfully make the run_model task fail: [link](https://screenshot.googleplex.com/4cguJZbvcBuHP8D)

# Checklist


xlml/apis/test_config.py
```diff
@@ -204,11 +204,11 @@ def benchmark_id(self) -> str:
 
   @property
   def setup_script(self) -> Optional[str]:
-    return '\n'.join(self.set_up_cmds)
+    return '\n'.join(('set -xue', *self.set_up_cmds))
```
will-cromar, line 207: What do you think of adding this to the SSH implementation instead? I'm thinking that if these run in a Docker container, they won't be guaranteed to run in `bash`. Within our SSH function, we can guarantee that

RissyRan: I don't have a strong preference on this. I will expect a new Test like `TpuVmTest(TestConfig[Tpu])` will be added for docker container, and we should have control there when calling `setup_script` or `run_script`. For SSH connection, we need to add logic into GPU and TPU implementation. So both strategies will add changes to multiple places.

will-cromar: Okay, let's try it this way. If needed, we can explicitly wrap the commands in `bash -cxue` to ensure the shell options are set.

will-cromar (approved):

xlml/apis/test_config.py
RissyRan: SGTM! Thanks!

PR #126, opened 2024-02-16
Add GPU clusters to deployments

will-cromar:
# Description

Add GPU cluster to deployment config.

# Tests

Applied config. Tested in PR that adds the tests: https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/111#issuecomment-1949326793

# Checklist

### DIFF
--- a/deployment/clusters.tf
+++ b/deployment/clusters.tf
@@ -1,5 +1,5 @@
 resource "google_container_cluster" "gpu-uc1" {
-  name     = "wcromar-test-cluster"
+  name     = "gpu-uc1"
   project  = "cloud-ml-auto-solutions"
   location = "us-central1"
 
@@ -30,9 +30,6 @@ resource "google_container_node_pool" "primary" {
     preemptible  = true
     machine_type = "e2-medium"
 
-    # Google recommends custom service accounts that have cloud-platform scope and permissions granted via IAM Roles.
-    # TODO: custom service account?
-    # service_account = google_service_account.default.email
     oauth_scopes    = [
       "https://www.googleapis.com/auth/cloud-platform"
     ]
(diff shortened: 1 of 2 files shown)


will-cromar:
Pulling this back to draft. I need to figure out how to deploy the headless service (for service discovery) from terraform.

RissyRan (commented):
Thanks Will! I don't have comment about Terraform template. Will those GPU capacity is taken once cluster is up? If so, @zpcore could you check if this change will affect your team's tests? I understand the resource request is painful.

will-cromar:
I've had a cluster of a the same size up for a few days for testing. Let me know if that caused a capacity crunch. I can switch this to a more conservative auto-scaling policy if needed.

zpcore:
Thanks for the head up. The GPU torchbench tests are scheduled to run at 3:00am PT. We will see tomorrow.

deployment/clusters.tf
```diff
@@ -0,0 +1,95 @@
...
+  host  = "https://${google_container_cluster.gpu-uc1.endpoint}"
+  token = data.google_client_config.provider.access_token
+  cluster_ca_certificate = base64decode(
+    google_container_cluster.gpu-uc1.master_auth[0].cluster_ca_certificate,
+  )
+}
+
+// Headless service required for service discovery within a Job.
+// Pods will be addressable as `hostname.headless-svc`
+resource "kubernetes_service" "example" {
```
vanbasten23, line 85: Does it mean it is a static and long lived service that we only run once?

will-cromar: Right. Every pod that defines the label `headless-svc=true` will register under the same service. 

I would have liked to create a service for each job that matches the first worker, but confusingly the job index is defined as an annotation and not a label. Services can only match labels.

vanbasten23: Got it. Each job has a unique name and they all share the same service (name is static).

PR #132, opened 2024-02-20
fix clean_up_gcs attribute issue

zpcore:
# Description

Fix the `clean_up_gcs` attribute not specified by the user in `task_metric_config`.

Add `if hasattr(task_metric_config, "clean_up_gcs")...` condition check in metric.py to avoid error out.

# Tests

See if other test can pass this time:

https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/maxtext_nightly/grid

### DIFF
--- a/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
+++ b/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
@@ -82,7 +82,8 @@ def get_torchbench_tpu_config(
 
   set_up_cmds = set_up_torchbench_tpu(model_name)
   local_output_location = "~/xla/benchmarks/output/metric_report.jsonl"
-  gcs_location = f"{gcs_bucket.BENCHMARK_OUTPUT_DIR}/torchbench_config/{tpu_version}/{int(datetime.datetime.now().timestamp())}/metric_report_tpu.jsonl"
+  gcs_location = f"{gcs_bucket.BENCHMARK_OUTPUT_DIR}/torchbench_config/{tpu_version}/metric_report_tpu.jsonl"
+
   if not model_name or model_name.lower() == "all":
     run_filter = " "
   else:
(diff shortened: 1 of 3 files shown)


xlml/utils/metric.py
```diff
@@ -165,6 +165,23 @@ def download_object_from_gcs(source_location: str, destination_location: str) ->
   logging.info(f"Download JSON Lines file to: {destination_location}")
 
 
+def delete_object_from_gcs(source_location: str) -> None:
```
RissyRan, line 168: Can we avoid to delete object from `metric.py`? This may affect other tests if we did not config it right.

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
```diff
@@ -221,7 +223,7 @@ def get_torchbench_gpu_config(
 
   set_up_cmds = set_up_torchbench_gpu(model_name, nvidia_driver_version)
   local_output_location = "/tmp/xla/benchmarks/output/metric_report.jsonl"
-  gcs_location = f"{gcs_bucket.BENCHMARK_OUTPUT_DIR}/torchbench_config/{accelerator_type}/{int(datetime.datetime.now().timestamp())}/metric_report_gpu.jsonl"
+  gcs_location = f"{gcs_bucket.BENCHMARK_OUTPUT_DIR}/torchbench_config/{accelerator_type}/metric_report_gpu.jsonl"
```
RissyRan, line 226: I think `{int(datetime.datetime.now().timestamp())}` can be fixed by a xcom or variable in Airflow. It may not be a good idea to delete your benchmarking report for previous runs.

xlml/utils/metric.py
zpcore: You need to specify `clean_up_gcs=True,` in the job config in order to delete the result object in GCS bucket.

The reason to add this is to prevent `post_process` step from using the result from previous run if the current run fail to generate result.

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
zpcore: I can change to use xcom. My only concern is that this will add an extra function entry in `task.py` like `run_with_gcs_name`. Which one do you prefer?

zpcore:
Point to a new PR: https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/133. We can drop this one.

PR #133, opened 2024-02-20
add variable to distinguist gcs result bucket

zpcore:
# Description

Use variable to identify the run id and use the id for gcs result folder.

# Tests

[maxtext_nightly run](https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/maxtext_nightly/grid)

[torchbench run](https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla-torchbench/grid)

### DIFF
--- a/dags/solutions_team/configs/pax/solutionsteam_pax_supported_config.py
+++ b/dags/solutions_team/configs/pax/solutionsteam_pax_supported_config.py
@@ -71,18 +71,13 @@ def get_setup_cmds(
     raise RuntimeError(f"Please specify set up cmds for: {pax_version.value}.")
 
 
-def get_runtime_version(pax_version: PaxVersion, tpu_version: TpuVersion) -> str:
+def get_runtime_version(tpu_version: TpuVersion) -> str:
   if tpu_version is TpuVersion.V5E:
     return RuntimeVersion.V2_ALPHA_TPUV5_LITE.value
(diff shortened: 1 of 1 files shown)


will-cromar (commented):
I'm not sure I understand the use of the Variable here. Isn't the random path something you could pass through XCOM?

In either case, you should generate that name in a dedicated task rather than as part of the GPU or TPU name.

zpcore:
> I'm not sure I understand the use of the Variable here. Isn't the random path something you could pass through XCOM?
> 
> In either case, you should generate that name in a dedicated task rather than as part of the GPU or TPU name.

I agree using the xcom. Generating the gcs name in TPU name function is workable but make it quite messy.

RissyRan:
I am fine with using either xcom or variable, but I have concern to generate this speicific ID through tpu/gpu name generation. The previous PR of TensorFlow is to fix TPU name inconsistency, so it makes sense to have that variable set in the name generation function. I think you should consider a helper function in your config to set this ID and then get it.

zpcore:
Let's use the PR here: https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/138. That one use XComarg and run the command by check env variable passed to the fabric ssh runner.

PR #138, opened 2024-02-22
support env to runner to dynamic change run command

zpcore:
# Description

1. Support env to runner to dynamic change run command
2. Support passing gs file location to `run_model` and `post_process` function
3. update `provision` phase to generate gcs file based on current timestamp.

Note:
You need to to set `MetricConfig.use_runtime_generated_filename=True` to tell post_process to use process generated filename.

# Tests

torchbench: [run](https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla-torchbench/grid)
Other's test like maxtext: [run](https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/maxtext_nightly/grid?dag_run_id=manual__2024-02-22T03%3A03%3A45.831358%2B00%3A00)
Flax [run](https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/flax_latest_supported/grid)

### DIFF
--- a/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
+++ b/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
@@ -82,7 +82,7 @@ def get_torchbench_tpu_config(
 
   set_up_cmds = set_up_torchbench_tpu(model_name)
   local_output_location = "~/xla/benchmarks/output/metric_report.jsonl"
-  # The gcs location can be overwrite if we use run_with_gcs_name_generatioin().
+  # The gcs location can be overwritten if we use run_with_gcs_name_generatioin().
   gcs_location = (
       f"{gcs_bucket.BENCHMARK_OUTPUT_DIR}/torchbench_config/metric_report_tpu.jsonl"
   )
@@ -224,7 +224,7 @@ def get_torchbench_gpu_config(
 
   set_up_cmds = set_up_torchbench_gpu(model_name, nvidia_driver_version)
   local_output_location = "/tmp/xla/benchmarks/output/metric_report.jsonl"
-  # The gcs location can be overwrite if we use run_with_gcs_name_generatioin().
+  # The gcs location can be overwritten if we use run_with_gcs_name_generatioin().
   gcs_location = (
       f"{gcs_bucket.BENCHMARK_OUTPUT_DIR}/torchbench_config/metric_report_gpu.jsonl"
   )
--- a/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
+++ b/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
@@ -95,7 +95,7 @@ def get_torchbench_tpu_config(
       ),
       "rm -rf ~/xla/benchmarks/output/metric_report.jsonl",
       "python ~/xla/benchmarks/result_analyzer.py --output-format=jsonl",
-      f"gsutil cp {local_output_location} ${metric_config.MetricPath.GCS_OUTPUT.value}",
+      f"gsutil cp {local_output_location} ${metric_config.SshEnvVars.GCS_OUTPUT.value}",
   )
 
   test_name = f"torchbench_{model_name}" if model_name else "torchbench_all"
@@ -240,7 +240,7 @@ def get_torchbench_gpu_config(
           "sudo docker cp $(sudo docker ps | awk 'NR==2 { print $1 }')"
           f":{local_output_location} ./"
       ),
-      f"gsutil cp metric_report.jsonl ${metric_config.MetricPath.GCS_OUTPUT.value}",
+      f"gsutil cp metric_report.jsonl ${metric_config.SshEnvVars.GCS_OUTPUT.value}",
   )
 
   test_name = f"torchbench_{model_name}" if model_name else "torchbench_all"
--- a/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
+++ b/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
@@ -98,7 +98,7 @@ def get_torchbench_tpu_config(
       ),
       "rm -rf ~/xla/benchmarks/output/metric_report.jsonl",
       "python ~/xla/benchmarks/result_analyzer.py --output-format=jsonl",
-      f"gsutil cp {local_output_location} $gcs_location",
+      f"gsutil cp {local_output_location} $OUTPUT_DIR",
   )
 
   test_name = f"torchbench_{model_name}" if model_name else "torchbench_all"
@@ -251,7 +251,7 @@ def get_torchbench_gpu_config(
           "sudo docker cp $(sudo docker ps | awk 'NR==2 { print $1 }')"
           f":{local_output_location} ./"
       ),
-      f"gsutil cp metric_report.jsonl $gcs_location",
+      f"gsutil cp metric_report.jsonl $OUTPUT_DIR",
   )
 
   test_name = f"torchbench_{model_name}" if model_name else "torchbench_all"
--- a/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
+++ b/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
@@ -82,7 +82,10 @@ def get_torchbench_tpu_config(
 
   set_up_cmds = set_up_torchbench_tpu(model_name)
   local_output_location = "~/xla/benchmarks/output/metric_report.jsonl"
-  gcs_location = f"{gcs_bucket.BENCHMARK_OUTPUT_DIR}/torchbench_config/{tpu_version}/{int(datetime.datetime.now().timestamp())}/metric_report_tpu.jsonl"
+  # The gcs location can be overwrite if we use run_with_gcs_name_generatioin().
+  gcs_location = (
+      f"{gcs_bucket.BENCHMARK_OUTPUT_DIR}/torchbench_config/metric_report_tpu.jsonl"
+  )
   if not model_name or model_name.lower() == "all":
     run_filter = " "
   else:
@@ -95,7 +98,7 @@ def get_torchbench_tpu_config(
       ),
       "rm -rf ~/xla/benchmarks/output/metric_report.jsonl",
       "python ~/xla/benchmarks/result_analyzer.py --output-format=jsonl",
-      f"gsutil cp {local_output_location} {gcs_location}",
+      f"gsutil cp {local_output_location} ${{gcs_location}}",
   )
 
   test_name = f"torchbench_{model_name}" if model_name else "torchbench_all"
@@ -221,7 +224,10 @@ def get_torchbench_gpu_config(
 
   set_up_cmds = set_up_torchbench_gpu(model_name, nvidia_driver_version)
   local_output_location = "/tmp/xla/benchmarks/output/metric_report.jsonl"
-  gcs_location = f"{gcs_bucket.BENCHMARK_OUTPUT_DIR}/torchbench_config/{accelerator_type}/{int(datetime.datetime.now().timestamp())}/metric_report_gpu.jsonl"
+  # The gcs location can be overwrite if we use run_with_gcs_name_generatioin().
+  gcs_location = (
+      f"{gcs_bucket.BENCHMARK_OUTPUT_DIR}/torchbench_config/metric_report_gpu.jsonl"
+  )
 
   if not model_name or model_name.lower() == "all":
     run_filter = " "
@@ -245,7 +251,7 @@ def get_torchbench_gpu_config(
           "sudo docker cp $(sudo docker ps | awk 'NR==2 { print $1 }')"
           f":{local_output_location} ./"
       ),
-      f"gsutil cp metric_report.jsonl {gcs_location}",
+      f"gsutil cp metric_report.jsonl ${{gcs_location}}",
   )
 
   test_name = f"torchbench_{model_name}" if model_name else "torchbench_all"
--- a/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
+++ b/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
@@ -82,10 +82,7 @@ def get_torchbench_tpu_config(
 
   set_up_cmds = set_up_torchbench_tpu(model_name)
   local_output_location = "~/xla/benchmarks/output/metric_report.jsonl"
-  # The gcs location can be overwritten if we use run_with_gcs_name_generatioin().
-  gcs_location = (
-      f"{gcs_bucket.BENCHMARK_OUTPUT_DIR}/torchbench_config/metric_report_tpu.jsonl"
-  )
+
   if not model_name or model_name.lower() == "all":
     run_filter = " "
   else:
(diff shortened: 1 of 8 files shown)


xlml/utils/metric.py
```diff
@@ -598,6 +598,7 @@ def process_metrics(
     task_metric_config: metric_config.MetricConfig,
     task_gcp_config: gcp_config.GCPConfig,
     use_startup_script: bool = False,
+    file_location: str = None,
```
RissyRan, line 601: Wondering if this is the right change to support this feature. We could have lots of environment variables in the future, and we cannot just adding more if/else here.

zpcore: Sad thing is that `task_metric_config` currently is decided at DAG creation time, it's hard to update the `task_metric_config.profile.file_locations`.  Let me do a try to see if I can overwrite the `task_metric_config` s.t. we can get rid of passing `file_location` here.

zpcore:
The DAG check failed due to `dict` type hint. The CI is using python<=3.8. I will update the hint keyword.
https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html

zpcore:
I tried with modifying `self.task_metric_config.json_lines.file_location` directly, but in post_process, it turns out `task_metric_config.json_lines.file_location` is still a `PlainXComArg` object instead of string. Below is the code I tested:
```
  def run_with_gcs_name_generation(self) -> DAGNode:
    with TaskGroup(
        group_id=self.task_test_config.benchmark_id, prefix_group_id=True
    ) as group:
      gcs_location = name_format.generate_gcs_file_location(
          self.task_test_config.benchmark_id
      )
      provision, ip_address, instance_name, ssh_keys = self.provision()
      run_model = self.run_model(ip_address, ssh_keys, {"gcs_location": gcs_location})
      self.task_metric_config.json_lines.file_location = gcs_location
      post_process = self.post_process()
      clean_up = self.clean_up(
          instance_name, self.task_gcp_config.project_name, self.task_gcp_config.zone
      )

      gcs_location >> provision >> run_model >> post_process >> clean_up

    return group
```

I am curious if this line https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/f511cd2b8a5066b2d6cc13c9c0a15bbe4f6987b8/xlml/apis/task.py#L102 returns a string as intended or the code is using the string format of `PlainXComArg` that nobody noticed.

xlml/apis/task.py
```diff
@@ -410,6 +433,24 @@ def run(self) -> DAGNode:
       provision >> run_model >> post_process >> clean_up
     return group
 
+  def run_with_gcs_name_generation(self) -> DAGNode:
+    with TaskGroup(
+        group_id=self.task_test_config.benchmark_id, prefix_group_id=True
+    ) as group:
+      gcs_location = name_format.generate_gcs_file_location(
```
will-cromar, line 440: nit: can `gcs_location` move into `provision`?

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
```diff
@@ -95,7 +98,7 @@ def get_torchbench_tpu_config(
       ),
       "rm -rf ~/xla/benchmarks/output/metric_report.jsonl",
       "python ~/xla/benchmarks/result_analyzer.py --output-format=jsonl",
-      f"gsutil cp {local_output_location} {gcs_location}",
+      f"gsutil cp {local_output_location} ${{gcs_location}}",
```
will-cromar, line 101: Can you clarify the double brace `${{` syntax here? Looking at the code, I think this will format to something like `${gs://.../torchbench_config/metric_report_tpu.jsonl}`, which would not be a valid shell variable

xlml/utils/gpu.py
```diff
@@ -350,13 +350,19 @@ def create_resource(
 
 
 @task
-def ssh_host(ip_address: str, cmds: Iterable[str], ssh_keys: ssh.SshKeys) -> None:
+def ssh_host(
+    ip_address: str,
+    cmds: Iterable[str],
+    ssh_keys: ssh.SshKeys,
+    env: dict[str, str] = None,
```
will-cromar, line 357: nit: should this also be `Dict` instead of `dict`?

xlml/apis/task.py
```diff
@@ -471,9 +513,10 @@ def run_model(
         resource,
         self.task_test_config.test_script,
         ssh_keys,
+        env,
     )
 
-  def post_process(self) -> DAGNode:
+  def post_process(self, result_file_location: airflow.XComArg = None) -> DAGNode:
```
will-cromar, line 519: nit: `Optional[...]` if `None` is a valid value

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
zpcore: This will be formatted to string like `gsutil cp /tmp/metric ${gcs_location}`, where it will use the env `${gcs_location}`.

xlml/apis/task.py
zpcore: Yes, the trade off is that we will introduce an extra return in `provision` function but can get rid of `gcs_location` stage. Are we fine with that?

zpcore: I think moving the `gcs_location` to `provision` stage makes the code clumpy:
```
  def provision(self) -> Tuple[DAGNode, airflow.XComArg, airflow.XComArg, airflow.XComAry]:
    with TaskGroup(group_id="provision") as group:
      with TaskGroup(group_id="initialize"):
        tpu_name = tpu.generate_tpu_name(
            self.task_test_config.benchmark_id, self.tpu_name_env_var
        )
        ssh_keys = ssh.generate_ssh_keys()

      queued_resource_op, queued_resource_name = tpu.create_queued_resource(
          tpu_name,
          self.task_gcp_config,
          ssh_keys,
          self.tpu_create_timeout,
          self.task_test_config,
      )

      gcs_location = name_format.generate_gcs_file_location(
          self.task_test_config.benchmark_id
      )

      queued_resource_op >> tpu.ssh_tpu.override(task_id="setup")(
          queued_resource_name,
          # TODO(wcromar): remove split
          self.task_test_config.setup_script,
          ssh_keys,
          self.all_workers,
      )

    return group, queued_resource_name, ssh_keys, gcs_location 
```
Not everyone wants the gcs_location here. We can do refactor later together with other customized `run_with...` functions.

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
will-cromar: Oh, I understand now. You're using an f-string for `{local_output_location}`, and `{{` escapes to `{` so you can reference the shell variable `${gcs_location}`.

Can you find another to format this string? Maybe just put `$gcs_location` to eliminate ambiguity around the braces

xlml/apis/task.py
will-cromar: Is there any harm in always generating a GCS location name? I actually think this would be helpful. The old system used to pass in a unique GCS `${MODEL_DIR}` in every case, and most (but not all) tests used it. This would let us cut out yet another variant of `run`

xlml/utils/name_format.py
```diff
@@ -39,3 +40,17 @@ def generate_tb_file_location(run_name: str, base_output_directory: str) -> str:
   return os.path.join(
       base_output_directory, run_name, "tensorboard", "events.out.tfevents.*"
   )
+
+
+@task
+def generate_gcs_file_location(benchmark_id) -> str:
```
will-cromar, line 46: nit: please annotate the type of `benchmark_id` too

xlml/apis/task.py
```diff
@@ -77,6 +77,24 @@ def run(self) -> DAGNode:
...
+  def run_with_gcs_name_generation(self) -> DAGNode:
+    with TaskGroup(
+        group_id=self.task_test_config.benchmark_id, prefix_group_id=True
+    ) as group:
+      gcs_location = name_format.generate_gcs_file_location(
+          self.task_test_config.benchmark_id
+      )
+      provision, queued_resource, ssh_keys = self.provision()
+      run_model = self.run_model(
+          queued_resource, ssh_keys, {"gcs_location": gcs_location}
```
will-cromar, line 89: ultra-nit: this variable becomes an env var, not a shell var, so please capitalize `GCS_LOCATION`

https://askubuntu.com/a/26322

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
zpcore: Right, `$gcs_location` should be simpler.

xlml/apis/task.py
zpcore: Oh, if we want to let user specify the gcs location as a standard routine, then adding this to provision should be good. Let me make the update.

xlml/apis/task.py
will-cromar: Now that I think about this, should we make this variable name more specific? We used to call this `MODEL_DIR` to match a convention in TF Estimator, but `OUTPUT_DIR` is clearer IMO

RissyRan (commented):
Sorry for the late reply! 

I was trying to see if we have a simpler solution for this. So I tried templates.

I updated the location in the config file to `gcs_location = f"{gcs_bucket.BENCHMARK_OUTPUT_DIR}/torchbench_config/{tpu_version}/ran/{{{{ts_nodash}}}}/metric_report_tpu.jsonl"`, and I am able to get file uploaded as expexcted https://screenshot.googleplex.com/3cfVgz6jfaafTMZ.

However, I cannot get this value in post_process method. It just gave me the string as `{{ts_nodash}}`: https://screenshot.googleplex.com/5EhKPkttZoFFEWe. I tried to add @task decorator, and it just skipped the method.

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
```diff
@@ -221,7 +224,10 @@ def get_torchbench_gpu_config(
 
   set_up_cmds = set_up_torchbench_gpu(model_name, nvidia_driver_version)
   local_output_location = "/tmp/xla/benchmarks/output/metric_report.jsonl"
-  gcs_location = f"{gcs_bucket.BENCHMARK_OUTPUT_DIR}/torchbench_config/{accelerator_type}/{int(datetime.datetime.now().timestamp())}/metric_report_gpu.jsonl"
+  # The gcs location can be overwritten if we use run_with_gcs_name_generatioin().
+  gcs_location = (
```
RissyRan, line 228: As we will directly use the passed in location in post_metric, do we still need this to construct the `metric_config.MetricConfig`. If not, can we remove them here?

xlml/apis/task.py
```diff
@@ -69,14 +69,30 @@ def run(self) -> DAGNode:
...
     return group
 
+  def run_with_gcs_name_generation(self) -> DAGNode:
+    with TaskGroup(
+        group_id=self.task_test_config.benchmark_id, prefix_group_id=True
+    ) as group:
+      
+      provision, queued_resource, ssh_keys, gcs_location = self.provision()
+      run_model = self.run_model(
+          queued_resource, ssh_keys, {"OUTPUT_DIR": gcs_location}
```
RissyRan, line 87: I see this `OUTPUT_DIR` used multiple places, can we have a enum variable somewhere to re-use to avoid typo?

xlml/apis/task.py
```diff
@@ -155,23 +171,25 @@ def provision(self) -> Tuple[DAGNode, airflow.XComArg, airflow.XComArg]:
...
+      )
       queued_resource_op, queued_resource_name = tpu.create_queued_resource(
           tpu_name,
           self.task_gcp_config,
           ssh_keys,
           self.tpu_create_timeout,
           self.task_test_config,
       )
-      queued_resource_op >> tpu.ssh_tpu.override(task_id="setup")(
+      output_location >> queued_resource_op >> tpu.ssh_tpu.override(task_id="setup")(
```
RissyRan, line 185: Will be a better logic if we put queued_resource_op >> metric_output_location >> setup? similar to GPU if apply

RissyRan (commented):
Thanks Pei! 

This is a big change, could you please have a few tests to ensure we are not breaking other team's tests on airflow - especially metric processing:
1) your torchbench test;
2) one simple test from flax;
3) one simple test from maxtext.

xlml/apis/task.py
```diff
@@ -69,14 +69,30 @@ def run(self) -> DAGNode:
     with TaskGroup(
         group_id=self.task_test_config.benchmark_id, prefix_group_id=True
     ) as group:
-      provision, queued_resource, ssh_keys = self.provision()
+      provision, queued_resource, ssh_keys, _ = self.provision()
```
will-cromar, line 72: Can you merge `run_with_gcs_name_generation` with `run` if the only difference is making the GCS path available? AFAICT `generate_gcs_file_location` doesn't do anything destructive, and downstream tasks can decide whether or not they need to write metrics to GCS

zpcore: If so, the env variable will always been passing to the run_model and provision will assume `gcs_location` is consumed:
```
run_model = self.run_model(
          queued_resource, ssh_keys, {f"{metric_config.MetricPath.GCS_OUTPUT}": gcs_location}
      )
```
Do we want to keep `run` and `run_with_gcs_name_generation` separate or we want to pass argument like `specify_gcs_location:bool` to `run()`?

will-cromar: Yeah, I think that's fine. Consumers that don't need the extra env won't use it

xlml/apis/task.py
will-cromar: The control dependency is added when the `output_location` is passed to `run_model`.

xlml/apis/task.py
```diff
@@ -155,23 +171,25 @@ def provision(self) -> Tuple[DAGNode, airflow.XComArg, airflow.XComArg]:
             self.task_test_config.benchmark_id, self.tpu_name_env_var
         )
         ssh_keys = ssh.generate_ssh_keys()
-
+      
+      output_location = name_format.generate_gcs_file_location(
```
will-cromar, line 175: Can you move this into the `initialize` subgroup? The other tasks are similarly generating random data that is used by the rest of the test

zpcore:
The failure in `maxtext` seems fine, you can see in production [run](https://46c56b977f2b44158661f1ef150119c7-dot-us-central1.composer.googleusercontent.com/dags/maxtext_nightly/grid?dag_run_id=manual__2024-02-24T00%3A33%3A27.094499%2B00%3A00&task_id=maxtext-nightly-4xv4-8.provision.setup&tab=logs), it also fails in setup with the same error.

xlml/apis/metric_config.py
```diff
@@ -37,6 +37,10 @@ class AggregationStrategy(enum.Enum):
   MEDIAN = enum.auto()
 
 
+class MetricPath(enum.Enum):
```
will-cromar, line 40: nit: let's make this more general (like `SshEnvVars`) and put any future common variables we add to the ssh environment here.

will-cromar (approved):
Thanks! This is going to be a really useful feature

miladm:
@zpcore are we ready to merge this PR?

zpcore:
> @zpcore are we ready to merge this PR?

Merged, this should fix torchbench automation result not found issue.

PR #143, opened 2024-02-23
Support running QueuedResource TPU tests from local workstation

will-cromar:
# Description

- Add env var option to use TPU external IPs instead of internal (default)
  - Also use proxy required on Cloudtops when external IPs are enabled
- Add `local-airflow.sh` script to perform setup steps automatically.
  - Creates a new local temporary directory for `$AIRFLOW_HOME`
  - Symlinks the given path into the new temporary airflow environment's `dag` directory, limiting the number of DAGs in the local instance. Parsing all of the DAGs locally is extremely slow.
  - Enables external IPs and deserialization for data classes
  - Uses local `xlml` library. Changes are reflected live in the local airflow instance
  - Cold start time for a new temporary instance is <1 minute
- Correct a type annotation

Future work:

- Enable more task types
- Smooth out the process of running a specific test. Currently I've been creating a temporary DAG with one task and triggering that myself from the UI. Ideally we want to be able to launch one specific test in one command.

# Tests

<details>

image

</details>

# Checklist


scripts/local-airflow.sh
```diff
@@ -0,0 +1,20 @@
...
+echo "Creating Airflow instance to run $target_dag..."
+
+set -xue
+
+# TODO(wcromar): enable editable installs of `xlml` instead
+export PYTHONPATH=$PWD
+export XLMLTEST_CONFIGS=$PWD/dags/jsonnet
+export XLMLTEST_SSH_EXTERNAL_IPS=1
+
+export AIRFLOW_HOME=$(mktemp -d)
```
RissyRan, line 12: What's the `-d` flag here?

RissyRan (approved):

RissyRan (approved):

scripts/local-airflow.sh
will-cromar: This makes a directory instead of a file.

PR #141, opened 2024-02-23
Add TPU cleanup DAG

RissyRan:
# Description

Noticed the QR quota is fully occupied - [here](https://screenshot.googleplex.com/3cPzChSmv7EmXj4). This is caused by manually clear tasks in dev env (or occasionally the QR is not deleted).

Add cleanup DAG:
* Add QR clean up, and run every 10min
* Will continue to add/refactor task to clean up idle TPUs in coming PR

# Tests

Please describe the tests that you ran on Cloud VM to verify changes.

**Instruction and/or command lines to reproduce your tests:** ...

**List links for your tests (use go/shortn-gen for any internal link):** ...
Upload to Airflow and test:
- Successfully deletion - [link](https://screenshot.googleplex.com/BpJiHkhRtuX3jHi)
- All clean-up - [link](https://screenshot.googleplex.com/6nATQLfLsHwa8Yz)

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run one-shot tests and provided workload links above if applicable. 
- [x] I have made or will make corresponding changes to the doc if needed.

PR #145, opened 2024-02-26
use dag sensor to replace gcp wait_for_operation

zpcore:
# Description

Use dag sensor to replace gcp wait_for_operation when create GPU resources.

# Tests
https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla-torchbench/grid?dag_run_id=manual__2024-02-26T00%3A32%3A27.591511%2B00%3A00&task_id=torchbench_BERT_pyotrch-nvidia-tesla-v100.run_model&tab=logs

The update only affect the test using GPU.


xlml/utils/gpu.py
```diff
@@ -304,49 +124,166 @@ def create_resource(
...
+    # request is complete. We intentionally sleep for 60s to wait for the ip address to be
+    # accessible.
+    time.sleep(60)
+    instance_client = compute_v1.InstancesClient()
+    instance = instance_client.get(project=project_id, zone=zone, instance=instance)
+    if len(instance.network_interfaces) > 1:
+      logging.warning(f"GPU instance {gpu_name} has more than one network interface.")
+    return instance.network_interfaces[0].network_i_p
+
+  with TaskGroup(group_id="create_resource") as tg:
```
will-cromar, line 276: Can you wrap `create_resource` in an `@task_group` instead?

zpcore: This is basically the task copy paste from https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/c1642e07dfba11953942d8c789ae9df6ffa8a002/xlml/utils/tpu.py#L182. I can update to @task_group. Let me update TPU/GPU both.

zpcore: I just updated the GPU here first to use task_group decorator. TPU update will be made in a separate PR in case anything goes wrong.

xlml/utils/gpu.py
```diff
@@ -304,48 +124,163 @@ def create_resource(
...
+            protected against deletion or not.
+    Returns:
+        Ip address of the instance object created.
+    """
+    machine_type = accelerator.machine_type
+    image = get_image_from_family(project=image_project, family=image_family)
+    disk_type = f"zones/{gcp.zone}/diskTypes/pd-ssd"
+    disks = [disk_from_image(disk_type, 100, True, image.self_link)]
+    metadata = create_metadata({
+        # "install-nvidia-driver": "True",
```
RissyRan, line 162: Do we want to use true or false? If false is enough, could we remove this comment for less confusion?

xlml/utils/gpu.py
```diff
@@ -304,48 +124,163 @@ def create_resource(
...
+    })
+
+    accelerators = [
+        compute_v1.AcceleratorConfig(
+            accelerator_count=accelerator.count,
+            accelerator_type=f"projects/{gcp.project_name}/zones/{gcp.zone}/acceleratorTypes/{accelerator.accelerator_type}",
+        )
+    ]
+    service_account = compute_v1.ServiceAccount(
+        # email = "cloud-auto-ml-solutions@google.com",
```
RissyRan, line 175: Do we need it?

xlml/utils/gpu.py
```diff
@@ -304,48 +124,163 @@ def create_resource(
...
+    request.instance_resource = instance
+
+    # Wait for the create operation to complete.
+    logging.info(f"Creating the {instance_name} instance in {zone}...")
+
+    operation = instance_client.insert(request=request)
+    return operation.name
+
+  @task.sensor(poke_interval=60, timeout=1800, mode="reschedule")
+  def wait_for_resource_creation(operation_name: airflow.XComArg):
```
RissyRan, line 239: For accelerator provision, can we get timeout from task, something like https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/task.py#L59 in TPU? So users can define it.

xlml/utils/gpu.py
```diff
@@ -379,7 +314,6 @@ def ssh_host(
   ssh_group.run(cmds, env=env)
 
 
-# TODO(piz): Check why sometime GPU instance doesn't get deleted.
 @task_group
```
RissyRan, line 317: Is this `TODO(piz): Check why sometime GPU instance doesn't get deleted.` fixed?

xlml/utils/gpu.py
```diff
@@ -304,48 +124,163 @@ def create_resource(
...
+    status = operation.status.name
+    if status in ("RUNNING", "PENDING"):
+      logging.info(f"Resource create status: {status}, {operation.status_message}")
+      return False
+    else:
+      if operation.error:
+        logging.error(
+            f"Error during resource creation: [Code: {operation.http_error_status_code}]: {operation.http_error_message}",
+        )
+        raise operation.exception() or RuntimeError(operation.http_error_message)
```
RissyRan, line 257: Have you tested this route? wondering the result for `raise operation.exception() or RuntimeError(operation.http_error_message)`

xlml/utils/gpu.py
```diff
@@ -304,48 +124,163 @@ def create_resource(
...
+      if operation.warnings:
```
RissyRan, line 258: Nit: elif instead of if?

RissyRan (commented):
It seems your test link has been modified? I see all of tests are empty.

xlml/utils/gpu.py
zpcore: Not yet, I haven't seen the case where request was sent but failed the response.

xlml/utils/gpu.py
zpcore: My previous finding was that the GPU instance was not got deleted when we manually clear the task, but it turns out TPU shows the same issue and this behavior is expected.

xlml/utils/gpu.py
zpcore: Leave it there just in case we want to switch the account under the project. I can remove it to clear the code.

xlml/utils/gpu.py
zpcore: Right, removed!

xlml/utils/gpu.py
zpcore: Added to be consistent.

zpcore:
> It seems your test link has been modified? I see all of tests are empty.

Please check the link again. Note that the test is currently fail at run_model due to the code check-in in python/xla. This is unrelated to this PR, which only affect the vm provision.

xlml/utils/gpu.py
RissyRan: We are adding those 2 PRs to clean up TPUs. Do you think you could add similar one to GPU? 

1) https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/141
2) https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/148

RissyRan: Not now, but in future PRs.

xlml/utils/gpu.py
RissyRan: If that's the case, how are we confident this statement is correct/appropriate here?

RissyRan (commented):
Thanks Pei! Test for [record](https://screenshot.googleplex.com/C8foENHnKL54wJp).

xlml/utils/gpu.py
zpcore: This was referred from example like https://github.com/GoogleCloudPlatform/python-docs-samples/blob/a585e56c69a0da7e64286c54f438cfd863a03270/compute/client_library/snippets/instances/stop.py#L63-L70.

xlml/utils/gpu.py
zpcore: Sure, I will add GPU in the dag job.

xlml/utils/gpu.py
RissyRan: Thanks Pei.

```
This method will raise the exception received from `operation.exception()`
        or RuntimeError if there is no exception set, but there is an `error_code`
        set for the `operation`.`
```

RissyRan (approved):

PR #147, opened 2024-02-27
[PT/XLA] Add many more ResNet50 tests

will-cromar:
# Description

Porting several existing variants of this test to the new system.

@JackCaoG let me know if there's another nightly variant that you want to add here.

# Tests

Checked DAG construction:

<details>

image

</details>

# Checklist

### DIFF
--- a/dags/pytorch_xla/pytorchxla_torchvision.py
+++ b/dags/pytorch_xla/pytorchxla_torchvision.py
@@ -63,9 +63,31 @@ with models.DAG(
       ),
       US_CENTRAL1_C,
   ).run()
-  resnet_v4_8 = task.TpuQueuedResourceTask(
+  resnet_v3_8_tests = [
+      task.TpuQueuedResourceTask(
+          test_config.JSonnetTpuVmTest.from_pytorch(test),
+          US_CENTRAL1_C,
+      ).run()
+      for test in (
+          "pt-nightly-resnet50-pjrt-fake-v3-8-1vm",
+          "pt-nightly-resnet50-pjrt-ddp-fake-v3-8-1vm",
+      )
+  ]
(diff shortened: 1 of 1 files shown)


dags/pytorch_xla/pytorchxla_torchvision.py
```diff
@@ -78,7 +100,8 @@
       US_EAST1_C,
   ).run()
 
-  mnist_v2_8 >> (resnet_v2_8, resnet_v4_8, resnet_v5lp_4)
+  mnist_v2_8 >> (resnet_v2_8, *resnet_v4_8_tests, resnet_v4_32, resnet_v5lp_4)
+  resnet_v2_8 >> resnet_v3_8_tests
 
   resnet_v100_2x2 = task.GpuGkeTask(
```
RissyRan, line 106: We probably want to add post_process sooner to multihost GPU tests, as our dashboard summary is missing the track of those [tests](http://go/ml-auto-solutions-dashboard).

RissyRan (approved):
Thanks Will! The screenshot is nice, but I will probably prefer to have a link of the test that could direct me to original tests (composer env).

dags/pytorch_xla/pytorchxla_torchvision.py
will-cromar: I can add this once I'm done porting tests for the 2.3 release.

will-cromar:
> Thanks Will! The screenshot is nice, but I will probably prefer to have a link of the test that could direct me to original tests (composer env).

Makes sense. I skipped actually running any of these in this case because there are too many and these tests have already been running in the old system for a while. We also know that other variants have been passing in airflow.

PR #160, opened 2024-02-28
Fixing tf-nightly tests

aman2930:
# Description
Fixing tf-nightly tests. These were failing due to following reasons:
1. `tf-nightly` tests are dependent on `tf-keras-nightly`. But installing `tf-keras-nightly` will override the TPU specific TensorFlow with non-TPU compatible TF. Fixed this by installing `tf-keras-nightly` without any dependencies.
2. Already existing `model` directory was causing failure in setup step. 

Test Screenshot: https://screenshot.googleplex.com/84WJfrgbkKN8Bkf

# Checklist

### DIFF
--- a/dags/solutions_team/configs/tensorflow/common.py
+++ b/dags/solutions_team/configs/tensorflow/common.py
@@ -55,7 +55,9 @@ FEATURE_TIMEOUT = {
 
 
 CMD_PRINT_TF_VERSION = "python3 -c \"import tensorflow; print('Running using TensorFlow Version: ' + tensorflow.__version__)\""
-CMD_INSTALL_KERAS_NIGHTLY = "pip install --upgrade --no-deps --force-reinstall tf-keras-nightly"
+CMD_INSTALL_KERAS_NIGHTLY = (
+    "pip install --upgrade --no-deps --force-reinstall tf-keras-nightly"
+)
 
 
 def set_up_se_nightly() -> Tuple[str]:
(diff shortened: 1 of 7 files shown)


dags/solutions_team/configs/tensorflow/common.py
```diff
@@ -108,10 +108,11 @@ def set_up_tensorflow_2_16_keras() -> Tuple[str]:
 def set_up_google_tensorflow_models() -> Tuple[str]:
   """Common set up for tensorflow models."""
   return (
-      "sudo mkdir -p /usr/share/tpu && cd /usr/share/tpu && git clone https://github.com/tensorflow/models.git",
+      "if [ ! -d \"/usr/share/tpu/models\" ]; then sudo mkdir -p /usr/share/tpu && cd /usr/share/tpu && git clone https://github.com/tensorflow/models.git; fi",
```
EricLeFort, line 111: Should we do a pull if the directory still exists to make sure it's up to date?

will-cromar (approved):

will-cromar:
Thanks!

dags/solutions_team/configs/tensorflow/common.py
aman2930: Pulling it again is causing error `git fatal detected dubious ownership in repository` which occurs because Sourcetree runs git with a different user. This might require extra steps to disable the ownership verification.

Also, I think pulling it again should not be necessary as this is the nightly directory where the models and the image both need to be in sync. When nightly image build, it syncs the same model directory at the same time. So it should be latest and up to date.

RissyRan (approved):

EricLeFort (approved):

PR #162, opened 2024-02-29
Add nightly AOT compilation GKE tests for MaxText configs

raymondzouu:
# Description

Adding nightly AOT compilation GKE tests for MaxText configs running on `v4-8-maxtext` cluster. Runs AOT scripts for all configs found in `v4`, `v5e`, and `v5p` directories in https://github.com/google/maxtext/tree/main/MaxText/configs

# Tests

Ran in local composer environment http://shortn/_6NgKh4BaUn. There is a failure with the 128B config on v5e-256 due to a compiler bug being tracked in b/327447977. This failure is unrelated to the set up of the nightly testing in XLML.

# Checklist

### DIFF
--- a/dags/multipod/maxtext_configs_aot.py
+++ b/dags/multipod/maxtext_configs_aot.py
@@ -1,0 +1,105 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+A DAG to run AOT compilation tests for MaxText model configs.
+"""
+import datetime
+from airflow import models
+from dags import composer_env, test_owner
+from dags.vm_resource import TpuVersion, Zone, DockerImage
+from dags.multipod.configs import gke_config
+from dags.multipod.configs.common import SetupMode
+
+# Run once a day at 6 am UTC (10 pm PST)
+SCHEDULED_TIME = "0 5 * * *" if composer_env.is_prod_env() else None
+
+with models.DAG(
+    dag_id="maxtext_configs_aot",
+    schedule=SCHEDULED_TIME,
+    tags=["multipod_team", "maxtext", "stable", "nightly"],
+    start_date=datetime.datetime(2024, 2, 19),
+    catchup=False,
+    concurrency=2,
+) as dag:
+  # Testing configurations
+  model_configs = {
+      # accelerator: [(model_size, num_cores), ...],
+      "v4": [("22b", 128), ("52b", 384)],
+      "v5e": [("16b", 256), ("32b", 256), ("64b", 256), ("128b", 256)],
+      "v5p": [
+          ("32b", 128),
+          ("64b", 128),
+          ("128b", 256),
+          ("128b", 512),
+          ("256b", 1024),
+          ("512b", 1024),
+          ("1024b", 2048),
+          ("1024b", 4096),
+      ],
+  }
+  num_slices = [1, 2]
+  docker_images = [
+      (SetupMode.STABLE, DockerImage.MAXTEXT_JAX_STABLE),
+      (SetupMode.NIGHTLY, DockerImage.MAXTEXT_JAX_NIGHTLY),
+  ]
+
+  run_model_cmds_dict = {}
+  for tpu, models in model_configs.items():
+    run_model_cmds = []
+    for model_size, num_cores in models:
+      for n in num_slices:
+        cmd = f"bash MaxText/configs/{tpu}/{model_size}.sh EXECUTABLE=train_compile.py M_COMPILE_TOPOLOGY={tpu}-{num_cores} M_COMPILE_TOPOLOGY_NUM_SLICES={n}"
+        run_model_cmds.append(cmd)
+    run_model_cmds_dict[tpu] = run_model_cmds
+
+  for mode, image in docker_images:
+    maxtext_v4_configs_test = gke_config.get_gke_config(
+        tpu_version=TpuVersion.V4,
+        tpu_cores=8,
+        tpu_zone=Zone.US_CENTRAL2_B.value,
+        time_out_in_min=60,
+        test_name=f"maxtext-aot-v4-{mode.value}",
+        run_model_cmds=run_model_cmds_dict["v4"],
+        docker_image=image.value,
+        test_owner=test_owner.RAYMOND_Z,
+    ).run()
+
+    maxtext_v5e_configs_test = gke_config.get_gke_config(
+        tpu_version=TpuVersion.V4,
+        tpu_cores=8,
+        tpu_zone=Zone.US_CENTRAL2_B.value,
+        time_out_in_min=60,
+        test_name=f"maxtext-aot-v5e-{mode.value}",
+        run_model_cmds=run_model_cmds_dict["v5e"],
+        docker_image=image.value,
+        test_owner=test_owner.RAYMOND_Z,
+    ).run()
+
+    maxtext_v5p_configs_test = gke_config.get_gke_config(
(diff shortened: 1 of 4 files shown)


gobbleturk (approved):
LGTM. We could run these on a CPU instead of a TPU as well

dags/vm_resource.py
```diff
@@ -98,11 +98,14 @@ class ClusterName(enum.Enum):
...
+  V4_8_MULTISLICE_CLUSTER = "v4-8-maxtext"
   V4_128_MULTISLICE_CLUSTER = "v4-bodaborg"
   V5E_16_MULTISLICE_CLUSTER = "v5e-16-bodaborg"
   V5E_256_MULTISLICE_CLUSTER = "v5e-256-bodaborg"
 
 
 class DockerImage(enum.Enum):
   XPK_JAX_TEST = "gcr.io/cloud-ml-auto-solutions/xpk_jax_test:latest"
+  MAXTEXT_JAX_STABLE = "gcr.io/tpu-prod-env-multipod/maxtext_jax_stable:latest"
+  MAXTEXT_JAX_NIGHTLY = "gcr.io/tpu-prod-env-multipod/maxtext_jax_nightly:latest"
```
jonb377, line 110: xpk won't pull the latest image if it's already on the node, so we still risk running stale images here. Since we'll be using the same tag `latest`, could you add a step to cache-image [here](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/utils/xpk.py#L68)? cc @Obliviour for semantics of cache-image.

Alternatively, we could tag the images with the build date and use the new tag for each run.

gobbleturk: Ah interesting, glad you thought of this. Currently the maxtext images are also tagged by date, we would just need be careful of timing - e.g. the date requested has already been pushed. For this reason I like using `latest` instead of date as long as we could guarantee not stale, but using date also seems fine

raymondzouu: I think there may be a limit to how many images can be cached cc @Obliviour to confirm. However, if there is not a limit then both methods sgtm.

gobbleturk: I may be misunderstanding here but would removing all docker images here on the test host like we do in [maxtext](https://github.com/google/maxtext/blob/2e7e96744826665680a41e81cd393cf2d58cada0/.github/workflows/UnitTests.yml#L63-L66) guarantee us to pull the latest gcr copy?

raymondzouu: I changed it to use the date tagged image for now, since we know for sure it will pull the fresh image.

dags/multipod/maxtext_configs_aot.py
```diff
@@ -0,0 +1,101 @@
...
+      (SetupMode.STABLE, DockerImage.MAXTEXT_JAX_STABLE),
+      (SetupMode.NIGHTLY, DockerImage.MAXTEXT_JAX_NIGHTLY),
+  ]
+
+  run_model_cmds_dict = {}
+  for tpu, models in model_configs.items():
+    run_model_cmds = []
+    for model_size, num_cores in models:
+      for n in num_slices:
+        cmd = f"bash MaxText/configs/{tpu}/{model_size}.sh EXECUTABLE=train_compile.py M_COMPILE_TOPOLOGY={tpu}-{num_cores} M_COMPILE_TOPOLOGY_NUM_SLICES={n}"
```
jonb377, line 63: Are these tests just to verify that compilation succeeds, or are there other metrics being generated?

dags/multipod/maxtext_configs_aot.py
```diff
@@ -0,0 +1,101 @@
...
+"""
+import datetime
+from airflow import models
+from dags import composer_env, test_owner
+from dags.vm_resource import TpuVersion, Zone, DockerImage
+from dags.multipod.configs import maxtext_gke_config
+from dags.multipod.configs.common import SetupMode
+
+# Run once a day at 6 am UTC (10 pm PST)
+SCHEDULED_TIME = "0 5 * * *" if composer_env.is_prod_env() else None
```
jonb377, line 26: 10pm should be fine, but for my understanding what time are the docker builds triggered?

dags/vm_resource.py
```diff
@@ -98,11 +99,14 @@ class ClusterName(enum.Enum):
...
+  V4_8_MULTISLICE_CLUSTER = "v4-8-maxtext"
   V4_128_MULTISLICE_CLUSTER = "v4-bodaborg"
   V5E_16_MULTISLICE_CLUSTER = "v5e-16-bodaborg"
   V5E_256_MULTISLICE_CLUSTER = "v5e-256-bodaborg"
 
 
 class DockerImage(enum.Enum):
   XPK_JAX_TEST = "gcr.io/cloud-ml-auto-solutions/xpk_jax_test:latest"
+  MAXTEXT_JAX_STABLE = f"gcr.io/tpu-prod-env-multipod/maxtext_jax_stable:{datetime.datetime.today().strftime('%Y-%m-%d')}"
+  MAXTEXT_JAX_NIGHTLY = f"gcr.io/tpu-prod-env-multipod/maxtext_jax_nightly:{datetime.datetime.today().strftime('%Y-%m-%d')}"
```
jonb377, line 111: My Airflow knowledge is pretty high-level, so just want to confirm - this gets evaluated at test execution time, not just at DAG build time, right?

jonb377 (approved):
Looks great, thanks Raymond!

dags/vm_resource.py
raymondzouu: I believe this actually gets evaluated at DAG build time cc @RissyRan to confirm. Therefore we should be careful using datetime in Airflow but I think in this case it is fine since we are just getting the date rather than the exact time.

dags/vm_resource.py
```diff
@@ -98,11 +99,14 @@ class ClusterName(enum.Enum):
   V4_32_CLUSTER = "mas-v4-32"
   V5E_4_CLUSTER = "mas-v5e-4"
   V5E_16_CLUSTER = "mas-v5e-16"
+  V4_8_MULTISLICE_CLUSTER = "v4-8-maxtext"
```
RissyRan, line 102: Could you also help add `v4-16-maxtext` for future cases?

dags/multipod/configs/maxtext_gke_config.py
```diff
@@ -0,0 +1,65 @@
+# Copyright 2023 Google LLC
```
RissyRan, line 1: nit: 2024

RissyRan (approved):

dags/multipod/configs/maxtext_gke_config.py
```diff
@@ -0,0 +1,65 @@
...
+
+"""Utilities to construct configs for maxtext DAG on GKE."""
+
+from xlml.apis import gcp_config, metric_config, task, test_config
+from dags import test_owner
+from dags.vm_resource import TpuVersion, Project, ClusterName
+from typing import Iterable
+
+
+def get_maxtext_configs_aot_config(
```
gobbleturk, line 23: Is there anything aot specific about this config? I would like to use it for my convergence tests - could this be named get_maxtext_gke_config instead?

gobbleturk (approved):

dags/multipod/configs/maxtext_gke_config.py
gobbleturk: Also nothing maxtext specific - could the file be renamed `gke_config` and the method `get_gke_config`?

dags/vm_resource.py
raymondzouu: Done!

dags/multipod/configs/maxtext_gke_config.py
raymondzouu: Good call, done!

gobbleturk: Sorry I already submitted https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/180 - I thought you were going to be out for a week! I renamed the method to `get_maxtext_gke_config`, but I think it would be better if you renamed the file and method to be maxtext-free (but keeping the name the same is fine too). I hope rebasing on top of my PR isn't too painful... I basically rebased on top of this PR except I changed the method name. If you do change the name you will also have to change in the [maxtext_convergence](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/dags/multipod/maxtext_convergence.py) test I added

raymondzouu: Sounds good! I rebased and changed the name to be maxtext-free

PR #163, opened 2024-02-29
Add Pylint check

RissyRan:
# Description

Add Pylint check:
* Considered both Pylint and Ruff. Decided to go with Pylint to match G3 style, as we will need to have a sub-repo for NPI workflow.
* Use google style [here](https://github.com/google/styleguide/blob/gh-pages/pylintrc), but I will need to update `max-line-length` to 125 instead of 80 to match Pyink format style.
* I addressed some straightforward errors, will need following PRs to address rest.
* Currently, set the threshold to 9/10 to gradually increase our code style quality (currently dags folder is 10/10, and xlml folder is 9.2/10).
* Add github action, and update doc and scripts accordingly.

# Tests

Test it locally and on PR checks.

All checks passed.

# Checklist


xlml/apis/test_config.py
```diff
@@ -135,7 +135,6 @@ class TestConfig(abc.ABC, Generic[A]):
   @abc.abstractmethod
   def benchmark_id(self) -> str:
     """Unique key for metrics generated by this test."""
-    ...
```
will-cromar, line 138: I assume pylint was complaining about an unused expression. Should this be `pass` or `raise NotImplementedError()` instead?

.github/workflows/pylint-check.yml
```diff
@@ -0,0 +1,30 @@
...
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pylint
+    - name: Analysing the code with pylint
+      run: |
+        pylint ./dags --rcfile=scripts/.pylintrc --fail-under=9
```
will-cromar, line 29: Can you run `scripts/code-style.sh` here instead and check for diffs? I'd also like to add JSonnet styling there too (#149)

will-cromar (commented):
> Use google style [here](https://github.com/google/styleguide/blob/gh-pages/pylintrc), but I will need to update max-line-length to 125 instead of 80 to match Pyink format style.

Can you change `pyink` to match the style guide instead?

xlml/apis/test_config.py
RissyRan: I will prefer `raise NotImplementedError()` here, especially for benchmark_id. Will update it accordingly.

RissyRan:
> > Use google style [here](https://github.com/google/styleguide/blob/gh-pages/pylintrc), but I will need to update max-line-length to 125 instead of 80 to match Pyink format style.
> 
> Can you change `pyink` to match the style guide instead?

Can I update that to 80 in coming PR? It will be lots of changes there.

.github/workflows/pylint-check.yml
RissyRan: Sure thing.

will-cromar (approved):

.github/workflows/pylint-check.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pylint
```
will-cromar, line 26: One more nit: can you lock this to a specific pylint version (at least a major version) so the check doesn't change when pylint updates?

RissyRan: Sure, updated to `pip install pylint==3.1.0`

PR #165, opened 2024-02-29
fix llama dynamo

zpcore:
# Description

Quick fix to fix llama2 dynamo script.

dynamo can't be set to True in the testing script:
```
python3 llama/example_text_completion.py True "/home/xl-ml-test/llama/7B" /home/xl-ml-test/spiece.model --max--max_gen_len=1000 --max_batch_size=2 --dynamo=True
```
Need to specify either openxla or openxla_eval.


will-cromar (approved):
Thanks!

vanbasten23 (approved):

PR #177, opened 2024-03-04
Catch 404 error when deleting QueuedResource

will-cromar:
# Description

If the TPU has already been deleted by the TPU cleanup job, we should not return an error status.

# Checklist


xlml/utils/tpu.py
```diff
@@ -239,7 +239,12 @@ def wait_for_tpu_deletion(qualified_name: str):
     creds, _ = google.auth.default()
     client = tpu_api.TpuClient(credentials=creds)
 
-    qr = client.get_queued_resource(name=qualified_name)
+    try:
+      qr = client.get_queued_resource(name=qualified_name)
+    except google.api_core.exceptions.NotFound:
+      logging.info(f'{qualified_name} was unexpectedly deleted')
```
RissyRan, line 245: Can we update to something like below? So we know one direction for debugging in the future.

```
f'{qualified_name} was deleted by clean up DAG or unexpectedly'
```

will-cromar: Done

RissyRan (approved):

zpcore:
Do you know why the resource deletion request can be lost during the clean up process?

will-cromar:
We don't actually monitor the individual delete operations, we monitor the parent QueuedResource to see if it goes into `SUSPENDED`. (There's actually some lag between when the nodes are all gone and we can delete the QR).

The TPU cleanup task deletes the QR separately sometimes, and then we get a 404 on `get_queued_resource`

PR #190, opened 2024-03-08
Allow running xpk workloads with standalone Airflow

jonb377:
# Description

Enable running XPK using a local Airflow environment, and integrate the `run_workload` using Taskflow. The task type should be conditionally determined by the execution environment: local Airflow tasks should use `task.docker`, and in the Composer environment `task.kubernetes` should be used.

*Note that the VM hosting the standalone Airflow server must have sufficient permissions to create workloads in the target GKE cluster.*

This change also adds a couple of ease-of-use enhancements for standalone runs:

- Disable login for the local server, and run on localhost
- Remove example DAGs to only show the desired one.

# Tests

- Ran example_gke_dag locally (1-slice failed due to VM permission issues, 2-slice started): http://screen/Aj5fJLtzuiCi9Dk
- Ran example_gke_dag in a Composer environment (manually killed the 4-slice run): http://shortn/_j9QVWlXXX1

# Checklist

### DIFF
--- a/deployment/cloud_composer_template.tf
+++ b/deployment/cloud_composer_template.tf
@@ -133,6 +133,7 @@ resource "google_composer_environment" "example_environment" {
       # Note: keep this in sync with .github/requirements.txt
       pypi_packages = {
         apache-airflow-providers-sendgrid = ""
+	apache-airflow-providers-docker   = ""
         fabric                            = ""
         google-cloud-tpu                  = ">=1.16.0"
         jsonlines                         = ""
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -58,12 +58,7 @@ for this purpose.
 
 To run a dag file in a temporary local environment, use `local-airflow.sh`. The script will symlink just the DAG provided to speed up parsing times.
 
-This requires Airflow to be installed locally. Ensure you have the appropriate Airflow version installed as defined in [cloud_composer_template.tf](/deployment/cloud_composer_template.tf#L126). For version 2.6.3:
-
-```
-VERSION=2.6.3
-pip install apache-airflow==$VERSION --constraint 'https://raw.githubusercontent.com/apache/airflow/constraints-${VERSION}/constraints-3.10.txt'
-```
+This requires Airflow to be installed locally. You can configure your local environment by running `pip install -r .github/requirements.txt`.
 
 To run the local environment, use the following commands:
 
@@ -80,6 +75,6 @@ If you're running a JSonnet-based test, run this each time any time the test cha
 scripts/gen-configs.sh
 ```
 
-Airflow will print a link to a local instance as well as a temporary admin password. From the UI, find your dag and run it manually.
+Airflow will print a link to a local instance. From the UI, find your dag and run it manually.
 
 This functionality is extremely experimental, and not all DAGs are expected to work with a local standalone server. Only the Airflow server runs locally. Tests will still run in the project defined in each DAG, so use this option with caution.
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -58,6 +58,15 @@ for this purpose.
 
 To run a dag file in a temporary local environment, use `local-airflow.sh`. The script will symlink just the DAG provided to speed up parsing times.
 
+This requires Airflow to be installed locally. Ensure you have the appropriate Airflow version installed as defined in [cloud_composer_template.tf](/deployment/cloud_composer_template.tf#L126). For version 2.6.3:
+
+```
+VERSION=2.6.3
+pip install apache-airflow==$VERSION --constraint 'https://raw.githubusercontent.com/apache/airflow/constraints-${VERSION}/constraints-3.10.txt'
+```
+
+To run the local environment, use the following commands:
+
 ```
 gcloud auth login --update-adc
 scripts/local-airflow.sh path/to/dag_file.py
--- a/scripts/local-airflow.sh
+++ b/scripts/local-airflow.sh
@@ -8,6 +8,7 @@ set -xue
 export PYTHONPATH=$PWD
 export XLMLTEST_CONFIGS=$PWD/dags/jsonnet
 export XLMLTEST_SSH_EXTERNAL_IPS=1
+export XLMLTEST_LOCAL_AIRFLOW=1
 
 export AIRFLOW_HOME=$(mktemp -d)
 
@@ -16,5 +17,11 @@ ln -s $PWD/$target_dag $AIRFLOW_HOME/dags/
 
 cd $AIRFLOW_HOME
 
+# Disable auth login
+echo "AUTH_ROLE_PUBLIC = 'Admin'" > webserver_config.py
+
 export AIRFLOW__CORE__ALLOWED_DESERIALIZATION_CLASSES_REGEXP='.*'
+export AIRFLOW__CORE__LOAD_EXAMPLES=False
+export AIRFLOW__WEBSERVER__WEB_SERVER_HOST=localhost
+export AIRFLOW__API__AUTH_BACKENDS=airflow.api.auth.backend.default
 airflow standalone
--- a/scripts/local-airflow.sh
+++ b/scripts/local-airflow.sh
@@ -20,6 +20,7 @@ cd $AIRFLOW_HOME
 # Disable auth login
 echo "AUTH_ROLE_PUBLIC = 'Admin'" > webserver_config.py
 
+# Configs from https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html
 export AIRFLOW__CORE__ALLOWED_DESERIALIZATION_CLASSES_REGEXP='.*'
 export AIRFLOW__CORE__LOAD_EXAMPLES=False
 export AIRFLOW__WEBSERVER__WEB_SERVER_HOST=localhost
--- a/xlml/utils/xpk.py
+++ b/xlml/utils/xpk.py
@@ -15,9 +15,9 @@
 """Utilities to run workloads with xpk (https://github.com/google/xpk)."""
 
 import uuid
+import os
 from absl import logging
 from airflow.decorators import task
-from airflow.providers.cncf.kubernetes.operators.kubernetes_pod import KubernetesPodOperator
 from kubernetes import client as k8s_client
 from kubernetes.client import models as k8s_models
 from xlml.utils import gke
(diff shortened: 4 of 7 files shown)


docs/contributing.md
```diff
@@ -58,6 +58,15 @@ for this purpose.
 
 To run a dag file in a temporary local environment, use `local-airflow.sh`. The script will symlink just the DAG provided to speed up parsing times.
 
+This requires Airflow to be installed locally. Ensure you have the appropriate Airflow version installed as defined in [cloud_composer_template.tf](/deployment/cloud_composer_template.tf#L126). For version 2.6.3:
+
+```
+VERSION=2.6.3
+pip install apache-airflow==$VERSION --constraint 'https://raw.githubusercontent.com/apache/airflow/constraints-${VERSION}/constraints-3.10.txt'
```
will-cromar, line 65: Did this actually work for you? When I tried it after installing the `requirements.txt`, I had multiple conflicts with this version specifically.

scripts/local-airflow.sh
```diff
@@ -16,5 +17,11 @@ ln -s $PWD/$target_dag $AIRFLOW_HOME/dags/
 
 cd $AIRFLOW_HOME
 
+# Disable auth login
+echo "AUTH_ROLE_PUBLIC = 'Admin'" > webserver_config.py
```
will-cromar, line 21: Good find! I was wondering how to do this. Can you remove the part of the `contributing` doc that mentions a temporary admin password?

scripts/local-airflow.sh
```diff
@@ -16,5 +17,11 @@ ln -s $PWD/$target_dag $AIRFLOW_HOME/dags/
...
+
 export AIRFLOW__CORE__ALLOWED_DESERIALIZATION_CLASSES_REGEXP='.*'
+export AIRFLOW__CORE__LOAD_EXAMPLES=False
```
will-cromar, line 24: Also a good find!

xlml/utils/xpk.py
```diff
@@ -43,15 +43,9 @@ def run_workload(
...
-
-  The reason to use KubernetesPodOperator instead of BashOperator is that
-  xpk must run with Python 3.10 or greater; however, the latest version in
-  Composer is Python 3.8, and it's non-trivial to upgrade it as the  Composer
-  uses docker images that bundle Airflow releases with Python and other
-  libraries.
-  """
+):
+  """Run workload through xpk tool."""
+  import subprocess
```
will-cromar, line 48: Can subprocess be imported at the global scope? If not, can you comment why?

xlml/utils/xpk.py
```diff
@@ -79,17 +73,20 @@ def run_workload(
...
-      task_id=task_id,
-      name=benchmark_id,
-      cmds=["/bin/bash", "-c"],
-      arguments=[";".join(cmds)],
+  subprocess.run(["bash", "-c", ";".join(cmds)], check=True)
+
+
+# To support local execution using `scripts/local-airflow.sh`, run using
+# task.docker in local environments. Otherwise, task.kubernetes should be used.
+if "XLMLTEST_LOCAL_AIRFLOW" in os.environ:
```
will-cromar, line 81: I have a design question: do we want to have one flag for local airflow that enables a bunch of features (this), or individual flags for each relevant feature (`XLMLTEST_SSH_EXTERNAL_IPS`)?

cc @RissyRan

scripts/local-airflow.sh
```diff
@@ -16,5 +17,11 @@ ln -s $PWD/$target_dag $AIRFLOW_HOME/dags/
...
+export AIRFLOW__WEBSERVER__WEB_SERVER_HOST=localhost
+export AIRFLOW__API__AUTH_BACKENDS=airflow.api.auth.backend.default
```
will-cromar, line 26: What does this do?

will-cromar (approved):
Great work! LGTM

docs/contributing.md
will-cromar: I don't think this option will work before airflow 2.8: `AIRFLOW__CORE__ALLOWED_DESERIALIZATION_CLASSES_REGEXP`

deployment/cloud_composer_template.tf
```diff
@@ -133,6 +133,7 @@ resource "google_composer_environment" "example_environment" {
       # Note: keep this in sync with .github/requirements.txt
       pypi_packages = {
         apache-airflow-providers-sendgrid = ""
+	apache-airflow-providers-docker   = ""
```
RissyRan, line 136: Could you remind me to deploy the change once you merged? Thanks!

docs/contributing.md
RissyRan: For airflow installation, do you think it's better we add it in [dev-requirement](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/38812677c999b438b3b6f0bb0bf15de8ee407253/.github/dev-requirements.txt)?

scripts/local-airflow.sh
RissyRan: Probably we could add this file as comment to this script? https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html

docs/contributing.md
jonb377: Yeah, this worked for me - I created a clean conda env with python 3.10 and ran this first, then `pip install -r .github/requirements.txt`. I'll try once more to verify if the order matters.

What does `AIRFLOW__CORE__ALLOWED_DESERIALIZATION_CLASSES_REGEXP` do?

xlml/utils/xpk.py
jonb377: Ah spot on! I'll add a comment. The `task.docker/kubernetes` decorators don't capture any scope beyond the method they decorate, so the imports must be local.

xlml/utils/xpk.py
jonb377: That's a good point, for this change I only need to know if we're local or not. If that's generally the case, I would be in favor of a single env var to denote execution environment (standalone or Composer).

scripts/local-airflow.sh
jonb377: Interestingly, even with `AUTH_ROLE_PUBLIC = 'Admin'` I wasn't able to view task logs without logging in, the GET requests hit 401 Unauthorized. [airflow.api.auth.backend.default](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#auth-backends) overrides the local auth backend to allow all requests, which fixes this.

xlml/utils/xpk.py
RissyRan: I think it's best practice for Airflow to import in a smaller scope for expensive dependency. We could do that as well in refactoring. 

```
Note that import statements also count as top-level code. So, if you have an import statement that takes a long time or the imported module itself executes code at the top-level, that can also impact the performance of the scheduler. The following example illustrates how to handle expensive imports.
```
from [here](https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html)

xlml/utils/xpk.py
RissyRan: One flag sounds better to me.

RissyRan (commented):
Thanks Jon for this!

docs/contributing.md
will-cromar: This is required for plumbing data classes (particularly the SSH keys for QR tests) between steps. 

It was added in 2.8: `AIRFLOW__CORE__ALLOWED_DESERIALIZATION_CLASSES_REGEXP`

Annoyingly, there was another flag that had the same behavior in prior versions:

- Old (regex syntax): https://airflow.apache.org/docs/apache-airflow/2.6.3/configurations-ref.html#allowed-deserialization-classes
- New (glob syntax): https://airflow.apache.org/docs/apache-airflow/2.8.2/configurations-ref.html#allowed-deserialization-classes

It's a breaking change in a patch version bump, so it's hard to reconcile.

jonb377: Okay, that makes sense. I'd like to keep the local airflow version the same as what runs in the composer env, since I expect people will just test their changes locally without a dev env going forward and subtle version differences could bite us if a broken change rolls through.

It seems like we can roll back to regex in `AIRFLOW__CORE__ALLOWED_DESERIALIZATION_CLASSES` for 2.6.3, and I can look into how to make the install compatible with `requirements.txt`. What do you think?

jonb377: One other question, which Airflow version are we running in GitHub Actions? Will we need to change it to 2.6.3?

xlml/utils/xpk.py
jonb377: Interesting! From what I can tell, the `task` bodies aren't evaluated until DAG execution time, so it makes sense that imports should be local.

docs/contributing.md
will-cromar: I think the CI is installing the current version. That's what I get with `pip install -r requirements.txt`

jonb377: I'll revert this to 2.8 locally for now.

xlml/utils/xpk.py
jonb377: We'll consolidate in a separate change.

xlml/utils/xpk.py
```diff
@@ -79,17 +71,20 @@ def run_workload(
...
+  run_workload = task.docker(_run_workload, image="python:3.10")
+else:
+  run_workload = task.kubernetes(
+      _run_workload,
       namespace="composer-user-workloads",
       image="python:3.10",
       config_file="/home/airflow/composer_kube_config",
       kubernetes_conn_id="kubernetes_default",
-      startup_timeout_seconds=startup_timeout,
-      owner=task_owner,
```
jonb377, line 92: cc @RissyRan, just want to verify that this is OK to omit. In XPK, workload creation is quick, so `startup_timeout_seconds` doesn't seem necessary. `task_owner` is applied at the TaskGroup level, so I think this is OK to omit. Let me know if I'm missing anything.

RissyRan: I recall I met issues to load the image, if that's not the case in multislice anymore, we could remove `startup_timeout_seconds` for now.

for task_owner, discussing offline.

jonb377: I'm able to set `owner` through an override. I'll update the PR. Thanks Ran!

xlml/utils/xpk.py
RissyRan: Thanks Jon!

RissyRan (approved):

deployment/cloud_composer_template.tf
jonb377: @RissyRan I'll merge now.

PR #191, opened 2024-03-08
Use fake token to bypass torchbench HF init check

zpcore:
# Description

Torchbench always check the token if we import models from hugging face: https://github.com/pytorch/benchmark/blob/8ffe57ecb5df6af5c458684f1834fbe78adfda00/torchbenchmark/util/framework/huggingface/model_factory.py#L191-L195. The token is not a must for public models. We use fake token to bypass torchbench HF init check.

Details: https://b.corp.google.com/issues/324078896#comment6.


zpcore:
This PR will not affect the DAG job. We are unable to run the test at this time due to the version issue between torch and torch_xla.

will-cromar (approved):

PR #179, opened 2024-03-04
Update MetricConfig to use relative GCS filepath combine with unique GCS folder path generated in Provision

zpcore:
# Description

Update from `generate_gcs_file_location` to `generate_gcs_folder_location`. `Provision` stage will generate the unique gcs folder name based on dag_id and date.

If we set `MetricConfig.use_runtime_generated_gcs_folder` to True, then the result file will be `os.path.join(gcs_folder, MetricConfig.json_lines)` and `os.path.join(gcs_folder, MetricConfig.tensorboard_summary)`.

# Tests
[torchbench BERT_pytorch](https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla-torchbench/grid)
[flax_latest_supported](https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/flax_latest_supported/grid)
[maxtext_end_to_end](https://08398f98ca104983a06a424b851a390d-dot-us-central1.composer.googleusercontent.com/dags/maxtext_end_to_end/grid)

# Checklist

Before submitting this PR, please make sure (put X in square brackets):
- [x] I have performed a self-review of my code.
- [x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [x] I have run one-shot tests and provided workload links above if applicable. 
- [x] I have made or will make corresponding changes to the doc if needed.

PR #192, opened 2024-03-12
Example migration of Multipod XL ML test

jonb377:
# Description

The multipod team is migrating tests from the old XL ML test infrastructure. This change provides two example migrations:
- `gpt1-like` to demonstrate the process to migrate custom test scripts out of the old repository
- `maxtext-decode` to show how to migrate a MaxText end_to_end test.

The test scripts which were hosted in the old repository will be moved into the `dags/multipod/legacy_tests` folder.

These tests will run using XPK, and this change includes some minor related changes:
- Allow up to 20 hours for `wait_for_workload_completion`. Since XPK includes a scheduler through Kueue, we should allow tests to be queued up from Airflow and to run as capacity becomes available in the XPK cluster.
- Immediately terminate `wait_for_workload_completion` without retry if the workload pod fails.

# Tests

Ran the tests locally (process_metrics is broken locally): screen/63JfR5FWFhiuGcE
Ran the tests in a development Composer environment: https://fc137785b6c943a4af82d0671bf0ebe5-dot-us-central1.composer.googleusercontent.com/dags/multipod_legacy_xlml/grid?dag_run_id=manual__2024-03-14T22%3A57%3A06.860898%2B00%3A00&tab=logs&task_id=gpt1-like-stable-2xv4-16.run_model._run_workload

# Checklist

### DIFF
--- a/dags/multipod/configs/legacy_unit_test.py
+++ b/dags/multipod/configs/legacy_unit_test.py
@@ -40,8 +40,9 @@ def get_legacy_unit_test_config(
 ) -> task.TpuXpkTask:
   """
   Run a legacy unit test script.
-  `script_to_copy` will be copied into the workload container, and `test_cmd`
-  will run in the same directory.
+  `script_to_copy` is a script in the `dags/multipod/legacy_tests` folder to be
+  copied into the workload container, and `test_cmd` will run with the script
+  in the working directory.
   """
   job_gcp_config = gcp_config.GCPConfig(
       project_name=project_name,
--- a/dags/multipod/configs/legacy_unit_test.py
+++ b/dags/multipod/configs/legacy_unit_test.py
@@ -32,7 +32,6 @@ def get_legacy_unit_test_config(
     tpu_zone: str,
     time_out_in_min: int,
     test_name: str,
-    test_mode: common.SetupMode,
     test_owner: str,
     docker_image: str,
     num_slices: int = 1,
--- a/dags/multipod/configs/legacy_unit_test.py
+++ b/dags/multipod/configs/legacy_unit_test.py
@@ -17,6 +17,7 @@
 import datetime
 import os
 from xlml.apis import gcp_config, metric_config, task, test_config
+from base64 import b64encode
 from collections.abc import Iterable
 from dags import test_owner
 from dags.multipod.configs import common
@@ -24,7 +25,8 @@ from dags.vm_resource import TpuVersion, Project, RuntimeVersion, ClusterName
 
 
 def get_legacy_unit_test_config(
-    test_cmd: Iterable[str],
+    script_to_copy: str,
+    test_cmd: Iterable,
     tpu_version: TpuVersion,
     tpu_cores: int,
     tpu_zone: str,
@@ -39,7 +41,8 @@ def get_legacy_unit_test_config(
 ) -> task.TpuXpkTask:
   """
   Run a legacy unit test script.
-  test_cmd will be executed within the `dags/multipod/legacy_tests` directory.
+  `script_to_copy` will be copied into the workload container, and `test_cmd`
+  will run in the same directory.
   """
   job_gcp_config = gcp_config.GCPConfig(
       project_name=project_name,
@@ -47,9 +50,15 @@ def get_legacy_unit_test_config(
       dataset_name=metric_config.DatasetOption.XLML_DATASET,
   )
 
+  unit_test_folder = os.environ.get(
+      'XLMLTEST_MULTIPOD_LEGACY_TEST_DIR',
+      '/home/airflow/gcs/dags/dags/multipod/legacy_tests',
+  )
+  with open(os.path.join(unit_test_folder, script_to_copy), 'rb') as f:
+    encoded_script = b64encode(f.read()).decode()
+
   run_model_cmds = (
-      f'git clone https://github.com/GoogleCloudPlatform/ml-auto-solutions',
-      'cd ml-auto-solutions/dags/multipod/legacy_tests',
+      f'echo {encoded_script} | base64 -d > {script_to_copy}',
       'export TPU_STDERR_LOG_LEVEL=0 TPU_MIN_LOG_LEVEL=0 JAX_USE_PJRT_C_API_ON_TPU=1 TF_CPP_MIN_LOG_LEVEL=0',
       *test_cmd,
   )
--- a/dags/multipod/configs/legacy_unit_test.py
+++ b/dags/multipod/configs/legacy_unit_test.py
@@ -48,7 +48,6 @@ def get_legacy_unit_test_config(
   )
 
   run_model_cmds = (
-      'set -xue',
       f'git clone https://github.com/GoogleCloudPlatform/ml-auto-solutions',
       'cd ml-auto-solutions/dags/multipod/legacy_tests',
       'export TPU_STDERR_LOG_LEVEL=0 TPU_MIN_LOG_LEVEL=0 JAX_USE_PJRT_C_API_ON_TPU=1 TF_CPP_MIN_LOG_LEVEL=0',
--- a/dags/multipod/configs/legacy_unit_test.py
+++ b/dags/multipod/configs/legacy_unit_test.py
@@ -1,0 +1,76 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Utilities to run legacy tests from the old infrastructure."""
+
+import datetime
+import os
+from xlml.apis import gcp_config, metric_config, task, test_config
+from collections.abc import Iterable
+from dags import test_owner
+from dags.multipod.configs import common
+from dags.vm_resource import TpuVersion, Project, RuntimeVersion, ClusterName
+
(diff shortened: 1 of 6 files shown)


dags/multipod/legacy_tests/gpt1-like.py
```diff
@@ -0,0 +1,199 @@
+#!/usr/bin/python3
```
jonb377, line 1: This file is copied directly from our old unit tests repository. The only changes are from `code-style.sh`

dags/multipod/legacy.py
```diff
@@ -0,0 +1,77 @@
...
+from dags.vm_resource import TpuVersion, Zone, Project, DockerImage, ClusterName
+from dags.multipod.configs import legacy_unit_test, gke_config
+from dags.multipod.configs.common import SetupMode, Platform
+
+# Run once a day at 9 am UTC (1 am PST)
+SCHEDULED_TIME = "0 9 * * *" if composer_env.is_prod_env() else None
+
+
+for test_mode in [SetupMode.NIGHTLY, SetupMode.STABLE]:
+  with models.DAG(
```
will-cromar, line 29: The decorator style might be better here since you have a DAG that takes one argument:

```
@dag(...)
def multipod_legacy_xlml(docker_image):
  ...

multipod_legacy_xlml(DockerImage.MAXTEXT_JAX_NIGHTLY)
multipod_legacy_xlml(DockerImage.MAXTEXT_JAX_STABLE)
```

dags/multipod/legacy.py
```diff
@@ -0,0 +1,77 @@
...
+      catchup=False,
+      concurrency=2,
+  ) as dag:
+    docker_image = (
+        DockerImage.MAXTEXT_JAX_NIGHTLY
+        if test_mode == SetupMode.NIGHTLY
+        else DockerImage.MAXTEXT_JAX_STABLE
+    )
+    current_time = datetime.datetime.now()
+    current_date = current_time.strftime("%Y-%m-%d")
```
will-cromar, line 43: This will be the DAG parsing time, not the test run time. Can you reuse @zpcore's solution in #179?

dags/multipod/legacy.py
jonb377: There are a couple of DAG parameters that depend on the test mode (name & tags). I'll see how the refactor looks, but it will probably still need a for loop over setup modes.

dags/multipod/legacy.py
jonb377: Removed it for now to unblock the migration, only some of the tests will require it, and it's better to tackle the XPK integration in a separate PR.

xlml/utils/xpk.py
```diff
@@ -120,7 +130,10 @@ def wait_for_workload_completion(
     if pod.status.phase in ["Pending", "Running"]:
       logging.info(f"One pod phase is: {pod.status.phase}")
       return False
-    elif pod.status.phase in ["Failed", "Unknown"]:
+    elif pod.status.phase == "Failed":
+      # Don't keep retrying if the pod has failed
+      raise AirflowFailException(f"Bad pod phase: {pod.status.phase}")
```
will-cromar, line 135: Does `AirflowFailException` prevent retries? That's good to know

dags/multipod/configs/legacy_unit_test.py
```diff
@@ -0,0 +1,76 @@
...
+  Run a legacy unit test script.
+  test_cmd will be executed within the `dags/multipod/legacy_tests` directory.
+  """
+  job_gcp_config = gcp_config.GCPConfig(
+      project_name=project_name,
+      zone=tpu_zone,
+      dataset_name=metric_config.DatasetOption.XLML_DATASET,
+  )
+
+  run_model_cmds = (
```
will-cromar, line 50: Should these be `set_up_cmds`?

will-cromar (approved):
LGTM. Thanks!

dags/multipod/configs/legacy_unit_test.py
jonb377: That's a good point. XPK doesn't support set_up_cmds at the moment, since there's no separation between setup and the actual test script, which is pretty confusing behavior. I'll fix this.

xlml/utils/xpk.py
jonb377: Yep! https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/exceptions/index.html#airflow.exceptions.AirflowFailException

We need this especially now that I've increased the XPK sensor timeout, otherwise it will keep querying the already-dead pods for the full timeout period.

dags/multipod/configs/legacy_unit_test.py
will-cromar: Looks like the TestConfig does expose a setup script already: https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/b2f99ebfbad40e08a7a06134a9f050c53dfbed9d/xlml/apis/test_config.py#L245-L247

dags/multipod/configs/legacy_unit_test.py
```diff
@@ -0,0 +1,75 @@
...
+      f'git clone https://github.com/GoogleCloudPlatform/ml-auto-solutions',
```
will-cromar, line 51: I'm a little concerned this method makes this type of test hard to test. Any changes have to get pushed to the repository before you can do a test run, so it would really be better if the test case file originated from the `dags` folder at runtime.

Could you instead embed the file in the setup commands?

dags/multipod/configs/legacy_unit_test.py
jonb377: XPK only accepts one workload command, and we're using the test script currently: https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/b2f99ebfbad40e08a7a06134a9f050c53dfbed9d/xlml/apis/task.py#L388

It would be nice to have a field which joins both setup and test commands.

dags/multipod/configs/legacy_unit_test.py
jonb377: Sadly it would be messy... I've thought about this some already:

-  `xpk` tests run in an isolated environment from the airflow worker, so there's no way to have direct access to `dags` or SCP it over to the workload.
- There can't be any newlines in the workload command, so we'd need to encode the test script to embed it in the workload command (messy). 
- `xpk` also doesn't support configmaps to communicate the script across
- We can't use the `base_docker_image` feature to bake it into the image, since docker build would fail within the composer env.

It's unfortunate, but this is the existing pattern in the internal repo as well. cc @Obliviour in case there's an XPK feature I'm overlooking.

will-cromar: > There can't be any newlines in the workload command, so we'd need to encode the test script to embed it in the workload command (messy).

Python has a built-in function to escape strings including quotes, newlines, etc: https://docs.python.org/3/library/shlex.html#shlex.quote

You should be able to use that unless the XPK script is doing something funky with the input.

jonb377: It looks like `shlex.quote` doesn't replace newlines (unless I'm missing something), so we'd have to manage that separately. It does easily work with base64 encoding, but it's pretty gross... https://paste.googleplex.com/6607797074001920

xlml/utils/xpk.py
RissyRan: Nice!

dags/multipod/legacy_tests/gpt1-like.py
RissyRan: Curious what's the resource? I thought your team's tests are all in Jsonnet

RissyRan: Also, do we want to add Copyright header? as a OS standard

dags/multipod/configs/legacy_unit_test.py
RissyRan: Yeah, I think original I only included run_model cmds as set_up cmds is inside of provision task group in GCE route and we don't have that here. I think we could make the refactor to make it reasonable in the future, since we already have diff routes, i.e. GKE via xpk, GKE via API call, GCE etc.

dags/multipod/legacy.py
```diff
@@ -0,0 +1,73 @@
...
+    # TODO(jonbolin): Example for legacy unit test migration - evaluate whether
+    # to remove gpt1-like tests once test migration is complete.
+    for n_slice in [1, 2]:
+      legacy_unit_test.get_legacy_unit_test_config(
+          test_cmd=("python3 gpt1-like.py",),
+          tpu_version=TpuVersion.V4,
+          tpu_cores=16,
+          tpu_zone=Zone.US_CENTRAL2_B.value,
+          time_out_in_min=60,
+          test_name=f"gpt1-like",
```
RissyRan, line 53: Nit: not need f string right?

RissyRan (commented):
A few comments, overall LGTM!

dags/multipod/legacy_tests/gpt1-like.py
jonb377: Good catch! I'll add that.

Our internal tests are defined in jsonnet, but they call into unit test scripts.

dags/multipod/configs/legacy_unit_test.py
will-cromar: How the setup command should be used is very context-specific. In GCE, it's either a metadata startup script or another SSH shell, but for GKE, the distinction isn't meaningful. In the multi-GPU test type, I just stuck them together (plus or minus some entrypoint hackery you can ignore) 

https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/d146d716c9eed97380afec3c2c894687faf1976f/xlml/apis/task.py#L674-L679

In general, we want to keep the "what" separate from the "how", so I think setup script is still useful, but the current structure of `Task` blurs the lines too much. We have some planned refactoring and can address this later.

jonb377:
Thanks @will-cromar  and @RissyRan for the reviews! I'll merge this and engage the test owners for the rest of the migration.

RissyRan (approved):

PR #194, opened 2024-03-13
Add metric collection to maxtext convergence test

raymondzouu:
# Description

Add metric collection into BigQuery for maxtext convergence test. Will be able to store metrics like final loss.
This PR needs https://github.com/google/maxtext/pull/512 to be submitted first into maxtext to prevent run_name conflicts.

# Tests

Ran maxtext_convergence test in local composer env http://shortn/_25kKsnd9y2 and was able to automatically collect metrics into BigQuery http://screen/8nLdcWaZAze5NC2.

# Checklist

### DIFF
--- a/dags/multipod/maxtext_convergence.py
+++ b/dags/multipod/maxtext_convergence.py
@@ -63,4 +63,6 @@ with models.DAG(
         run_model_cmds=run_command,
         docker_image=DockerImage.MAXTEXT_JAX_STABLE.value,
         test_owner=test_owner.MATT_D,
-    ).run()
+        base_output_directory=base_output_directory,
+        metric_aggregation_strategy=metric_config.AggregationStrategy.LAST,
+    ).run_with_run_name_generation()
--- a/dags/multipod/configs/gke_config.py
+++ b/dags/multipod/configs/gke_config.py
@@ -36,6 +36,8 @@ def get_gke_config(
     dataset_name: metric_config.DatasetOption = metric_config.DatasetOption.XLML_DATASET,
     dataset_project: str = Project.CLOUD_ML_AUTO_SOLUTIONS.value,
     composer_project: str = Project.CLOUD_ML_AUTO_SOLUTIONS.value,
+    base_output_directory: str = None,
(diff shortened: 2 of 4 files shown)


gobbleturk (approved):
Thank you for adding metric collections for this test!

dags/multipod/maxtext_convergence.py
```diff
@@ -46,21 +46,31 @@
...
-        tpu_zone=Zone.US_CENTRAL2_B.value,
-        cluster_name=ClusterName.V4_128_MULTISLICE_CLUSTER.value,
-        time_out_in_min=300,
-        test_name=test_name,
-        run_model_cmds=run_command,
-        docker_image=DockerImage.MAXTEXT_JAX_STABLE.value,
-        test_owner=test_owner.MATT_D,
-    ).run()
+    maxtext_v4_configs_test = (
+        maxtext_sweep_gke_config.get_maxtext_sweep_gke_config(
```
jonb377, line 57: Just curious why we need the sweep?

jonb377 (approved):

dags/multipod/maxtext_convergence.py
raymondzouu: The sweep isn't needed here, just reusing the metric_config set up. Would it be better to modify `get_gke_config` instead so there's less confusion?

jonb377: Ahhh I see... Yeah, we can add it into `get_gke_config` if it's easy, but we should keep it as an optional config.

One small callout - `base_output_directory` is generated at dag parse time, and now that we share it between tasks it could change during DAG execution. Not a huge issue because this test runs early in the morning UTC and the format is using just the date, but this is something we'll need to address later with generated GCS paths.

raymondzouu: Done! Added the `metric_config` in `get_gke_config` only when the base_output_directory and aggregation_strategy are specified.

Agreed, that is a good callout. We'll definitely need a more unified and robust way for generating values that contain dates and time

PR #199, opened 2024-03-14
Nina/Add MaxText GPU tests with XPK

NinaCai:
# Description

Add decode test for Llama2, Mistral, and Gamma for GPU by using end-to-end bash scripts in MaxText repo via XPK.

# Tests

Upload to Airflow, and test.
https://540cef12d3da42ce97b48a97401b6c83-dot-us-central1.composer.googleusercontent.com/dags/maxtext_end_to_end/grid?dag_run_id=manual__2024-03-14T18%3A45%3A52.666342%2B00%3A00&task_id=maxtext-stable-test_llama2-h100-80gb-8.run_model._run_workload&tab=logs

# Checklist


dags/multipod/configs/gke_config.py
```diff
@@ -128,3 +129,52 @@ def get_gke_maxtext_nightly_config(
...
+    project_name: str = Project.GPU_PROD_ENV_MULTIPOD.value,
+    runtime_version: str = RuntimeVersion.TPU_UBUNTU2204_BASE.value,
+) -> task.GpuCreateResourceTask:
+  job_gcp_config = gcp_config.GCPConfig(
+      project_name=project_name,
+      zone=gpu_zone,
+      dataset_name=metric_config.DatasetOption.XLML_DATASET,
+  )
+
+  test_platform = common.Platform.GCE
```
jonb377, line 156: Should be GKE, right?

dags/vm_resource.py
```diff
@@ -31,6 +31,7 @@ class Project(enum.Enum):
   TPU_PROD_ENV_MULTIPOD = "tpu-prod-env-multipod"
   TPU_PROD_ENV_AUTOMATED = "tpu-prod-env-automated"
   CLOUD_TPU_MULTIPOD_DEV = "cloud-tpu-multipod-dev"
+  GPU_PROD_ENV_MULTIPOD = "supercomputer-testing"
```
jonb377, line 34: `SUPERCOMPUTER_TESTING = "supercomputer-testing"`

xlml/apis/task.py
```diff
@@ -418,6 +418,116 @@ def post_process(self) -> DAGNode:
       return group
 
 
+@dataclasses.dataclass
+class GpuXpkTask(BaseTask):
```
jonb377, line 422: Is there a functional difference between this and TpuXpkTask, or could we just make it a common Task between TPU and GPU?

xlml/apis/test_config.py
```diff
@@ -263,6 +263,41 @@ def _load_compiled_jsonnet(test_name: str) -> Any:
   return test
 
 
+@attrs.define
+class GpuGkeTest(TestConfig[Gpu]):
```
jonb377, line 267: Same here - just wondering if we can consolidate with the TpuGkeTest config. cc @will-cromar

dags/multipod/maxtext_end_to_end.py
```diff
@@ -42,9 +43,12 @@
       "gpt3": ["test_gpt3"],
   }
 
+  maxtext_end2end_tpu = [DummyOperator(task_id="maxtext_end2end_tpu")]
+  maxtext_end2end_gpu = [DummyOperator(task_id="maxtext_end2end_gpu")]
```
jonb377, line 47: Do we need to chain all GPU tests serially? This will slow down overall DAG execution, since the tests will execute one-at-a-time. With XPK, I'd prefer to kick off many tests at once and let Kueue handle the scheduling, and in Airflow just wait for the tests to finish.

dags/multipod/configs/gke_config.py
```diff
@@ -128,3 +129,52 @@ def get_gke_maxtext_nightly_config(
...
+  set_up_cmds = common.setup_maxtext(test_mode, test_platform)
+  run_model_cmds = (f"cd /tmp/maxtext && bash end_to_end/{test_script}.sh",)
+
+  job_test_config = test_config.GpuGkeTest(
+      test_config.Gpu(
+          machine_type=machine_type.value,
+          image_family=image_family.value,
```
jonb377, line 163: Are these fields needed in XPK?

jonb377 (commented):
Looking great, thanks Nina!

dags/multipod/maxtext_end_to_end.py
RissyRan: Yeah, I agree here. Previously we are using GCE resources (VMs), and does make sense to chain all tests if resources are insufficient. XPK works as a scheduling/queuing system. Unless, resources are quite tight in XPK, and those tests are timing out.

dags/vm_resource.py
```diff
@@ -92,7 +93,7 @@ class GpuVersion(enum.Enum):
 
   L4 = "nvidia-l4"
   A100 = "nvidia-tesla-a100"
-  H100 = "nvidia-h100-80gb"
+  H100 = "h100-80gb-8"
```
RissyRan, line 96: I think PyTorch/XLA team is using H100, has the name changed? If not, we may need another entry, i.e. XPK_H100

cc @zpcore

dags/vm_resource.py
```diff
@@ -133,3 +136,5 @@ class DockerImage(enum.Enum):
   XPK_MAXTEXT_TEST = "gcr.io/tpu-prod-env-multipod/xpk_maxtext_test:latest"
   MAXTEXT_JAX_STABLE = f"gcr.io/tpu-prod-env-multipod/maxtext_jax_stable:{datetime.datetime.today().strftime('%Y-%m-%d')}"
   MAXTEXT_JAX_NIGHTLY = f"gcr.io/tpu-prod-env-multipod/maxtext_jax_nightly:{datetime.datetime.today().strftime('%Y-%m-%d')}"
+  MAXTEXT_GPU_JAX_STABLE = f"gcr.io/tpu-prod-env-multipod/maxtext_gpu_jax_stable:{datetime.datetime.today().strftime('%Y-%m-%d')}"
```
RissyRan, line 139: Hi @jonb377, GCR is going to deprecate soon, does your team plan to migrate those to AR soon?

xlml/apis/task.py
RissyRan: Good point! I think has some small params differences, i.e. in TPU we pass `tpu-type`, but for CPU and GPU, we pass `device-type`. Let's unify them as one `XpkTask`?

xlml/apis/test_config.py
RissyRan: Will and I are thinking about refactor on those in coming weeks. Personally, I am fine with this part. Glad to hear more suggestions

RissyRan (commented):
Thanks Nina!

A few high-level comments/questions:
1) Is that ok to add title to something like `Add MaxText GPU tests with XPK`?
2) Is there any needed permission change in project `cloud-ml-auto-solutions`? If so, we could add them into the terraform template.

xlml/apis/task.py
jonb377: Nice, `device-type` will work for TPUs as well.

dags/vm_resource.py
jonb377: I saw that as well, it looks like it's not going to shut down support for projects already using GCR: https://cloud.google.com/container-registry/docs/deprecations/container-registry-deprecation

NinaCai:
> Thanks Nina!
> 
> A few high-level comments/questions:
> 
> 1. Is that ok to add title to something like `Add MaxText GPU tests with XPK`?
> 2. Is there any needed permission change in project `cloud-ml-auto-solutions`? If so, we could add them into the terraform template.

Thanks, @RissyRan, done for point 1. 
For point 2, cloud-ml-auto-solutions needs permission to use supercomputer-testing clusters.

dags/vm_resource.py
zpcore: `h100-80gb-8` is the machine type to host those GPUs while `nvidia-h100-80gb` is the GPU. We don't need to make the change.

zpcore: I didn't find `h100-80gb-8` is mentioned anywhere. Is this a typo?

NinaCai: > I didn't find `h100-80gb-8` is mentioned anywhere. Is this a typo?

This is the keyword used in xpk. I will change the name to `XPK_H100`.

dags/multipod/configs/gke_config.py
NinaCai: You are right, it is not needed

dags/vm_resource.py
```diff
@@ -133,3 +137,5 @@ class DockerImage(enum.Enum):
   XPK_MAXTEXT_TEST = "gcr.io/tpu-prod-env-multipod/xpk_maxtext_test:latest"
   MAXTEXT_JAX_STABLE = f"gcr.io/tpu-prod-env-multipod/maxtext_jax_stable:{datetime.datetime.today().strftime('%Y-%m-%d')}"
```
RissyRan, line 138: Nit: should we update this to `MAXTEXT_JAX_TPU_STABLE` then? same for `MAXTEXT_JAX_NIGHTLY`

xlml/utils/xpk.py
```diff
@@ -62,7 +62,7 @@ def _run_workload(
       (
           "python3 xpk.py workload create"
           f" --cluster={cluster_name} --workload={workload_id}"
-          f" --command='{run_cmds}' --tpu-type={accelerator_type}"
+          f" --command='{run_cmds}' --device-type={accelerator_type}"
```
RissyRan, line 65: This will be very tricky. I recall TPU will use `tpu-type`, and both GPU and CPU will use `device-type`. I wish this could be handled in xpk, so that we can only use `device-type` here.

xlml/apis/task.py
```diff
@@ -308,16 +308,16 @@ def clean_up(self, queued_resource: airflow.XComArg) -> DAGNode:
...
   Attributes:
-    task_test_config: Test configs to run on this TPU.
-    task_gcp_config: Runtime TPU creation parameters.
+    task_test_config: Test configs to run on this TPU/GPU.
+    task_gcp_config: Runtime TPU/GPU creation parameters.
     task_metric_config: Metric configs to process metrics.
   """
 
-  task_test_config: test_config.TestConfig[test_config.Tpu]
+  task_test_config: test_config.TestConfig[None]
```
RissyRan, line 320: Will it work use `test_config.Accelerator`?

RissyRan (commented):
Thanks Nina!

xlml/utils/xpk.py
NinaCai: I think `device-type` can be used for tpu, gpu and cpu. https://github.com/google/xpk/blob/948a57a87cecfa61d668f5bf7b8c30b2c893df56/xpk.py#L4290

RissyRan (commented):
It seems the test does not run properly? One example [here](https://screenshot.googleplex.com/4DeUczXEB2CZzcN)

xlml/apis/task.py
NinaCai: Changing to Accelerator has the error `type 'Accelerator' is not subscriptable`

NinaCai:
> It seems the test does not run properly? One example [here](https://screenshot.googleplex.com/4DeUczXEB2CZzcN)

I think it is just logging. The job exits normally (https://screenshot.googleplex.com/77oWiuuDoqj2WDo). 
Sometimes if the job is sent via xpk successfully, but fails execution, I can see a red dot at wait_for_workload_completion. ([example](https://screenshot.googleplex.com/hori9PdpNcHikNe))

xlml/utils/xpk.py
Obliviour: +1 to what @NinaCai said. We plan to depreciate tpu-type one day but didn't want to break current users. device-type supports all accelerators.

xlml/apis/test_config.py
will-cromar: I agree, I'm fine with this part in this PR, but this will get refactored fairly soon. We have a poor separation between the "what" and a "how" of the test. Any test that runs on GPU in docker should be using the same class or at least interface, and the particulars (like joining the commands with a semicolon to hack around `xpk`'s whitespace handling) should be part of the `Task`. (Also we'll rename `Task` :sweat_smile:). Stuff like `num_slices` should be part of the `accelerator`.

No action is required here IMO, unless rolling these two classes is trivial.

xlml/apis/test_config.py
```diff
@@ -263,6 +263,41 @@ def _load_compiled_jsonnet(test_name: str) -> Any:
...
+    num_slices: Number of GPU slices.
+  """
+
+  test_name: str
+  cluster_name: str
+  docker_image: str
+  set_up_cmds: Iterable[str]
+  run_model_cmds: Iterable[str]
+  startup_time_out_in_sec: int = attrs.field(default=300, kw_only=True)
+  num_slices: int = attrs.field(default=1, kw_only=True)
```
will-cromar, line 286: Does the term `num_slices` make sense here? Is this the number of hosts?

xlml/apis/task.py
```diff
@@ -308,16 +308,16 @@ def clean_up(self, queued_resource: airflow.XComArg) -> DAGNode:
 
 
 @dataclasses.dataclass
-class TpuXpkTask(BaseTask):
-  """This is a class to set up tasks for TPU provisioned by XPK tool.
+class XpkTask(BaseTask):
```
will-cromar, line 311: Nice, I like that you can combine the both XPK test types here.

xlml/apis/task.py
```diff
@@ -308,16 +308,16 @@ def clean_up(self, queued_resource: airflow.XComArg) -> DAGNode:
...
   Attributes:
-    task_test_config: Test configs to run on this TPU.
-    task_gcp_config: Runtime TPU creation parameters.
+    task_test_config: Test configs to run on this TPU/GPU.
+    task_gcp_config: Runtime TPU/GPU creation parameters.
     task_metric_config: Metric configs to process metrics.
   """
 
-  task_test_config: test_config.TestConfig[test_config.Tpu]
+  task_test_config: test_config.Accelerator
```
will-cromar, line 320: This type annotation is wrong. We don't have a type checker enabled right now, so I'm just eyeballing this. The only two test types that break here are `TpuGkeTest` and `GpuGkeTest`, right? In that case, the annotation would be `Union[TpuGkeTest, GpuGkeTest]`.

xlml/apis/test_config.py
will-cromar: micro-nit on the name, actually: this is the second GPU test type that runs on GKE (see `JSonnetGpuTest`). @NinaCai would you mind renaming this to `GpuXpkTest` to make it more specific?

will-cromar (approved):
Overall LGTM.

Please rename `GpuGkeTest`. I could go either way on renaming `TpuGkeTest` in this PR -- as mentioned in the comments, we have plans to clean up this code anyway.

xlml/apis/test_config.py
NinaCai: Yes, it is the number of nodes used for the workload.

xlml/apis/task.py
NinaCai: Thank you Will, just to be clear, I should change this line to `task_test_config: Union[test_config.TpuGkeTest, test_config.GpuXpkTest]` right?

will-cromar: Yeah

jonb377 (approved):
LGTM, thanks Nina!

RissyRan:
> > It seems the test does not run properly? One example [here](https://screenshot.googleplex.com/4DeUczXEB2CZzcN)
> 
> I think it is just logging. The job exits normally (https://screenshot.googleplex.com/77oWiuuDoqj2WDo). Sometimes if the job is sent via xpk successfully, but fails execution, I can see a red dot at wait_for_workload_completion. ([example](https://screenshot.googleplex.com/hori9PdpNcHikNe))

Interesting! I did not see training logs on GPU.

One example:
* TPU logs on Mistral: https://screenshot.googleplex.com/7wrNoL7yN6L8rTH
* GPU logs on Mistral: https://screenshot.googleplex.com/8jj9aPQRe2sxebP

Did I miss anything? 

cc @jonb377

jonb377:
Good catch @RissyRan, I see the XPK workload is running `(cd /deps && bash gpu_multi_process_run.sh)` instead of the train script. @NinaCai I think it's a bug in XPK: http://shortn/_nwVkm1F5sA

NinaCai:
> Good catch @RissyRan, I see the XPK workload is running `(cd /deps && bash gpu_multi_process_run.sh)` instead of the train script. @NinaCai I think it's a bug in XPK? http://shortn/_nwVkm1F5sA

This is on purpose, it just runs the commands in multi-processes. But the script is not in maxtext master branch, that is why it skips the whole script and just exits. @yangyuwei@google.com

PR #201, opened 2024-03-15
copy test scripts from release tag

lsy323:
# Description

Copy test scripts from release tag. Before this PR, the tests are copied from `master` branch, which may not be compatible with the r2.3 branch.

# Tests

Done locally.

# Checklist

### DIFF
--- a/dags/legacy_test/tests/pytorch/r2.3/common.libsonnet
+++ b/dags/legacy_test/tests/pytorch/r2.3/common.libsonnet
@@ -104,8 +104,7 @@ local volumes = import 'templates/volumes.libsonnet';
         pip install pillow
         git clone --depth=1 https://github.com/pytorch/pytorch.git
         cd pytorch
-        git clone https://github.com/pytorch/xla.git
-        cd xla && git checkout v2.3.0-rc2
+        git clone -b v2.3.0-rc2 https://github.com/pytorch/xla.git
       |||,
     },
     podTemplate+:: {
@@ -146,8 +145,7 @@ local volumes = import 'templates/volumes.libsonnet';
         pip install https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-2.3.0rc2-cp310-cp310-linux_x86_64.whl
 
         mkdir -p pytorch/xla
-        git clone https://github.com/pytorch/xla.git pytorch/xla
-        cd pytorch/xla && git checkout v2.3.0-rc2 && cd ../../
+        git clone -b v2.3.0-rc2 https://github.com/pytorch/xla.git pytorch/xla
 
         %s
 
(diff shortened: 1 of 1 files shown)


zpcore:
Thanks for catching the issue!

dags/legacy_test/tests/pytorch/r2.3/common.libsonnet
```diff
@@ -105,6 +105,7 @@ local volumes = import 'templates/volumes.libsonnet';
         git clone --depth=1 https://github.com/pytorch/pytorch.git
         cd pytorch
         git clone https://github.com/pytorch/xla.git
+        cd xla && git checkout v2.3.0-rc2
```
JackCaoG, line 108: do you plan to update this tag for every rc?

lsy323: We have to, the rc-x name is hard coded. I believe it's done in previous release too

lsy323: @will-cromar i think we should pull the rc tag instead of the release branch because the whl is based on tag instead of the branch. There is chance of having incompatibility between those two, when we have cherry picks after the latest tag is created.

will-cromar: Incompatibility between an RC in a test really shouldn't happen after the branch cut (since the API is stable after that point), unless there's an actual error in the test. We should pick up test fixes immediately rather than waiting for the next RC

lsy323: > Incompatibility between an RC in a test really shouldn't happen after the branch cut.

There is no guarantee on this. The purpose of this test flow is to test the rc whl, so I think we should use the test from rc tag. If we want to capture any failures in the release branch, we should have the a test targeting at the release branch HEAD.

lsy323: We need to update the test file anyway, need to change the url of the whl when the next rc is coming up.

lsy323: @will-cromar Thank you for the reminder of `git checkout -b`, never knew this before, updated the script so it's cleaner. (no `cd ../../`)

will-cromar: The only part of the test case that actually comes from this clone is the test file itself is the test file itself such as `test_operations` and `test_train_mp_imagenet`, which is not part of the release candidate. IIRC `-b` actually literally means `branch` in this case, so I don't think git will let you check out a tag directly :upside_down_face: 

Approving because you'll have to send a PR to update the other `rcX` strings for this release.

will-cromar (approved):

PR #204, opened 2024-03-15
stop running for Sunday and Mondays

zpcore:
# Description

Update to the torchbench schedule. This is no need to run on Sunday/Monday 3:00am PST. We will use the schedule to run other tasks to prevent resource competition.


JackCaoG:
Pytorch still have pr merged in during the weekend, I felt like there are still values running them during weekend.

zpcore:
> Pytorch still have pr merged in during the weekend, I felt like there are still values running them during weekend.

Oh, if this is the case, let's keep the current schedule.

PR #205, opened 2024-03-16
Refactored 2.16 release into a more reusable template

EricLeFort:
# Description

Refactors the release DAG to allow for easy swapping of versions for reuse in future releases and renames the modules to be more generic (`...tf_release...` instead of `...tf_2_16...`).

Also changes some typing references (e.g. `list` instead of `typing.List`) to adhere to Python best practices [PEP 585](https://peps.python.org/pep-0585/). This specific change will require python 3.11+ but adding `from __future__ import annotations` can allow python 3.7+. If we still need to support the older versions, I can update this change to include that.

# Tests

Ran 2.16 release PJRT and SE keras tests, both were successful.

Variety of tests of both nightly and 2.15.1 DAGS as well as from resnet and tf-keras tests:
http://shortn/_2sfrmqsyfr
http://shortn/_v5h49gyH6t
http://shortn/_MNIr72f9Z5
http://shortn/_WBsR2NNTlT
http://shortn/_MbmkG5jqOF
http://shortn/_HVtxmvAvoP
http://shortn/_XneLXMOyTn
Note: nightly tests are failing *but* they're failing for the same reason as they are in the dashboard.

# Checklist

### DIFF
--- a/dags/solutions_team/configs/tensorflow/solutionsteam_tf_release_supported_config.py
+++ b/dags/solutions_team/configs/tensorflow/solutionsteam_tf_release_supported_config.py
@@ -14,12 +14,23 @@
 
 """Utilities to construct configs for solutionsteam_tf_nightly_supported DAG."""
 
+from __future__ import annotations
+
 from xlml.apis import gcp_config, metric_config, task, test_config
 from dags import gcs_bucket, test_owner
 from dags.solutions_team.configs.tensorflow import common
 from airflow.models import Variable
 from dags.vm_resource import TpuVersion, Project, RuntimeVersion
-from typing import List
+
+
+MAJOR_VERSION = "2"
+MINOR_VERSION = "15"
+PATCH_VERSION = "1"
+LIBTPU_VERSION = "1.9.0"
+KERAS_VERSION = "2.15.1"
+MODELS_BRANCH = "r2.15.0"
+
+GS_VERSION_STR = f"tf-{MAJOR_VERSION}-{MINOR_VERSION}-{PATCH_VERSION}"
 
 
 def get_tf_keras_config(
@@ -42,10 +53,13 @@ def get_tf_keras_config(
       dataset_name=metric_config.DatasetOption.XLML_DATASET,
   )
 
-  set_up_cmds = common.install_tf_2_16() + common.set_up_tensorflow_2_16_keras()
+  set_up_cmds = common.set_up_keras(KERAS_VERSION)
+  set_up_cmds += common.install_tf(
+      MAJOR_VERSION, MINOR_VERSION, PATCH_VERSION, LIBTPU_VERSION
+  )
   if not is_pjrt and is_pod:
     set_up_cmds += common.set_up_se_nightly()
-  keras_test_name = f"tf_2_16_keras_api_{test_name}"
+  keras_test_name = f"tf_{MAJOR_VERSION}_{MINOR_VERSION}_keras_api_{test_name}"
   benchmark_id = f"{keras_test_name}-v{tpu_version.value}-{tpu_cores}"
   # Add default_var to pass DAG check
   # TODO(ranran): replace Variable.get() to XCOM when it applies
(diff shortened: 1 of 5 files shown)


EricLeFort:
Looks like GitHub's dag checker is using Python < 3.11 so I may as well just add the 3.7+ support. Fixing.

chandrasekhard2 (approved):

dags/solutions_team/configs/tensorflow/solutionsteam_tf_release_supported_config.py
```diff
@@ -14,12 +14,51 @@
...
+
 from xlml.apis import gcp_config, metric_config, task, test_config
 from dags import gcs_bucket, test_owner
 from dags.solutions_team.configs.tensorflow import common
 from airflow.models import Variable
 from dags.vm_resource import TpuVersion, Project, RuntimeVersion
-from typing import List
+
+
+MAJOR_VERSION = "2"
```
RissyRan, line 26: I feel this part can be passed in from dag as params for `install_tf()`, `set_up_keras()`, and `set_up_tensorflow_models()`? So those methods could actually moved to `common.py`? 

```
MAJOR_VERSION = "2"
MINOR_VERSION = "16"
PATCH_VERSION = "1"
LIBTPU_VERSION = "1.10.0"
KERAS_VERSION = "2.16.0rc0"
MODELS_BRANCH = "r2.16.0"
```

dags/solutions_team/solutionsteam_tf_release_se_supported.py
```diff
@@ -18,17 +18,19 @@
...
 
 
 # Release tests only need to run once, they can be run manually as needed
 SCHEDULED_TIME = None
+VERSION = f"{tf_config.MAJOR_VERSION}.{tf_config.MINOR_VERSION}"
+
 
 with models.DAG(
-    dag_id="tf_2_16_se_nightly_supported",
+    dag_id=f"tf_{tf_config.MAJOR_VERSION}_{tf_config.MINOR_VERSION}_se_nightly_supported",
```
RissyRan, line 31: I think major version and other variables could be defined in the dag level. In this case:

* 2.16 dag could pass in 2.16 as major version;
* 2.15.1 dag could pass in 2.15 as major version, and 1 as minor version

HDYT?

dags/solutions_team/configs/tensorflow/solutionsteam_tf_release_supported_config.py
EricLeFort: That's a good point, could add a line or two in the common versions of those functions rather than completely replicating them. I'll see what I can do.

dags/solutions_team/solutionsteam_tf_release_se_supported.py
EricLeFort: That would affect non-tf DAGS though which is a pretty big increase in scope for this change. I'm not sure I want to add it here?

I agree that being able to specify the version in the dag itself could be useful though! But perhaps as a future change.

EricLeFort: I think we'd only ever want to have a single release in flight at any given time -- that's what we do today anyway.

Although I could see a world where we leave release DAGs in place for future reference. I don't think it's a bad refactor regardless, it separates the logic of how to build a release DAG from which release to build a DAG for.

RissyRan:
Good point Eric! Do we need to actually monitor each release for 3-6
months? I think that's what we did last year unless the guidelines changed.

EricLeFort:
@RissyRan The 2.15.0 release DAG didn't even exist when I started on 2.16.0. We also only run release tests using manual triggers these days. So I'm not sure but it doesn't seem like it.

RissyRan (approved):
LGTM. I am not sure the current release process (if a single release still is monitored for 3-6 months, so we will have multiple release dags). 

I see owner has approved. We could always come back to update variables

dags/solutions_team/solutionsteam_tf_release_se_supported.py
EricLeFort: Spoke to Ran offline and we're going to add this as a future change. Resolving for now.

PR #208, opened 2024-03-19
Various XPK improvements

jonb377:
# Description

XPK has an additional layer of scheduling below Airflow. Jobs submitted can wait in a queue before execution. This change separates the timeout for the workload being scheduled from the workload's execution.

Since XPK workloads can be long-running (e.g. 5+ hour timeout for convergence tests), the tasks are kept as sensors instead of streaming the logs. This avoids consuming an Airflow worker for the duration of the test. Upon completion of the workload, the logs will be printed.

A few XPK-specific changes:
- The `run_workload` task is wrapped in a new TaskGroup, `launch_workload`, which handles creating the workload and waiting for it to be scheduled. This allows separating the timeouts for workload scheduling from execution.
- Dump the workload logs and link at the end of the `wait_for_workload_completion` task.
- Remove containerization required to launch the XPK workload, since the Composer environment now has python 3.11 which XPK supports.
- Support GCS path generation.

# Tests

Ran the new workflow locally: https://screenshot.googleplex.com/7AZCDggSYXrGe2z
- v4-8 launch failed due to my local VM permissions against the cluster, logs are correctly shown: https://screenshot.googleplex.com/7ogAMWziSdoqVjZ

Deployed to Composer environment: https://fc137785b6c943a4af82d0671bf0ebe5-dot-us-central1.composer.googleusercontent.com/dags/gke_example_dag/grid?tab=graph

# Checklist

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -65,6 +65,20 @@ scripts/local-airflow.sh path/to/dag_file.py
 
 Comment out any test cases in the DAG that you do not want to run, or create a temporary DAG file to avoid running all tests.
 
+
+##### XPK-based tests
+
+When running Airflow locally, XPK requires kubectl to support GKE-based authentication. Install the required component using
+
+```
+gcloud components install gke-gcloud-auth-plugin
+```
+
+If you encounter an error related to the gcloud installation being `managed by an external package manager`, you'll need to reinstall. The simplest way is to follow https://cloud.google.com/sdk/docs/downloads-interactive.
+
+
+##### JSonnet-based tests
+
 If you're running a JSonnet-based test, run this each time any time the test changes:
 
 ```
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -68,7 +68,10 @@ Comment out any test cases in the DAG that you do not want to run, or create a t
 
 ##### XPK-based tests
 
-When running Airflow locally, XPK requires kubectl to support GKE-based authentication. Install the required component using
+XPK will run in the same environment as the local airflow execution, and there are two XPK requirements to be aware of:
+
+1. Python version >= 3.10.
+1. kubectl must support GKE-based authentication. Install the required component using
 
 ```
 gcloud components install gke-gcloud-auth-plugin
--- a/xlml/apis/task.py
+++ b/xlml/apis/task.py
@@ -377,7 +377,10 @@ class XpkTask(BaseTask):
     """
     with TaskGroup(group_id="run_model") as group:
       workload_id = xpk.generate_workload_id(self.task_test_config.benchmark_id)
-      launch_workload = self.launch_workload(workload_id)
+      gcs_path = name_format.generate_gcs_folder_location(
+          self.task_test_config.benchmark_id
+      )
+      launch_workload = self.launch_workload(workload_id, gcs_path)
       wait_for_workload_completion = xpk.wait_for_workload_completion.override(
           timeout=self.task_test_config.time_out_in_min * 60,
       )(
@@ -387,10 +390,10 @@ class XpkTask(BaseTask):
           cluster_name=self.task_test_config.cluster_name,
       )
 
-      workload_id >> launch_workload >> wait_for_workload_completion
+      (workload_id, gcs_path) >> launch_workload >> wait_for_workload_completion
       return group
 
-  def launch_workload(self, workload_id: str) -> DAGNode:
+  def launch_workload(self, workload_id: str, gcs_path: str) -> DAGNode:
     """Create the workload and wait for it to provision."""
     with TaskGroup(group_id="launch_workload") as group:
       run_workload = xpk.run_workload.override(
@@ -402,6 +405,7 @@ class XpkTask(BaseTask):
           cluster_name=self.task_test_config.cluster_name,
           benchmark_id=self.task_test_config.benchmark_id,
           workload_id=workload_id,
+          gcs_path=gcs_path,
           docker_image=self.task_test_config.docker_image,
           accelerator_type=self.task_test_config.accelerator.name,
           run_cmds=self.task_test_config.test_script,
--- a/xlml/apis/task.py
+++ b/xlml/apis/task.py
@@ -320,6 +320,9 @@ class XpkTask(BaseTask):
   task_test_config: Union[test_config.TpuGkeTest, test_config.GpuXpkTest]
   task_gcp_config: gcp_config.GCPConfig
   task_metric_config: Optional[metric_config.MetricConfig] = None
+  workload_provision_timeout: datetime.timedelta = datetime.timedelta(
+      minutes=60
+  )
 
   def run(self) -> DAGNode:
     """Run a test job within a docker image.
@@ -388,6 +391,22 @@ class XpkTask(BaseTask):
           run_cmds=self.task_test_config.test_script,
           num_slices=self.task_test_config.num_slices,
       )
+      monitor_workload = self.monitor_workload(workload_id)
+
+      workload_id >> run_workload >> monitor_workload
+      return group
+
+  def monitor_workload(self, workload_id: str) -> DAGNode:
+    """Watch for the workload to provision and execute."""
+    with TaskGroup(group_id="monitor_workload") as group:
+      wait_for_workload_start = xpk.wait_for_workload_start.override(
+          timeout=self.workload_provision_timeout.total_seconds()
+      )(
+          workload_id=workload_id,
+          project_id=self.task_gcp_config.project_name,
+          region=self.task_gcp_config.zone[:-2],
+          cluster_name=self.task_test_config.cluster_name,
+      )
       wait_for_workload_completion = xpk.wait_for_workload_completion.override(
           timeout=self.task_test_config.time_out_in_min * 60,
       )(
@@ -396,8 +415,7 @@ class XpkTask(BaseTask):
           region=self.task_gcp_config.zone[:-2],
           cluster_name=self.task_test_config.cluster_name,
       )
-
-      workload_id >> run_workload >> wait_for_workload_completion
+      wait_for_workload_start >> wait_for_workload_completion
       return group
 
   def post_process(self) -> DAGNode:
--- a/xlml/apis/task.py
+++ b/xlml/apis/task.py
@@ -321,7 +321,7 @@ class XpkTask(BaseTask):
   task_gcp_config: gcp_config.GCPConfig
   task_metric_config: Optional[metric_config.MetricConfig] = None
   workload_provision_timeout: datetime.timedelta = datetime.timedelta(
-      minutes=60
+      minutes=300
   )
 
   def run(self) -> DAGNode:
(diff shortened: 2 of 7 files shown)


xlml/apis/task.py
```diff
@@ -320,6 +320,9 @@ class TpuXpkTask(BaseTask):
   task_test_config: test_config.TestConfig[test_config.Tpu]
   task_gcp_config: gcp_config.GCPConfig
   task_metric_config: Optional[metric_config.MetricConfig] = None
+  workload_provision_timeout: datetime.timedelta = datetime.timedelta(
```
RissyRan, line 323: Do we want to set 5 hours as default? or 20hours? so users do not need to specify it

RissyRan (commented):
Thanks Jon! LGTM! Just have a few nits.

xlml/utils/xpk.py
```diff
@@ -89,46 +89,73 @@ def _run_workload(
...
+    for pod in pods.items:
+      if pod.status.phase == "Failed":
+        # Don't keep retrying if the pod has failed
+        raise AirflowFailException(f"Bad pod phase: {pod.status.phase}")
+      elif pod.status.phase in ["Unknown"]:
+        raise RuntimeError(f"Bad pod phase: {pod.status.phase}")
+  finally:
+    # Print the logs of the last pod checked - either the first failed pod or
+    # the last successful one.
+    logs = core_api.read_namespaced_pod_log(
```
RissyRan, line 153: Do you have an example I could take a quick look?

xlml/apis/task.py
jonb377: Good point, let me increase it here. I think 5 hours is more reasonable, I was a bit overly ambitious to set 20 hours originally 😆

xlml/utils/xpk.py
jonb377: I'll deploy to the dev env with a successful run and one with a bad workload to show the failure case.

jonb377: Here is a link to a successful run: https://fc137785b6c943a4af82d0671bf0ebe5-dot-us-central1.composer.googleusercontent.com/dags/gke_example_dag/grid?root=&dag_run_id=manual__2024-03-26T19%3A31%3A59.601549%2B00%3A00&task_id=resnet-single-slice-v4-8.run_model.wait_for_workload_completion&tab=logs

I'll follow up with the failure case.

xlml/utils/xpk.py
```diff
@@ -37,98 +40,122 @@ def generate_workload_id(benchmark_id: str) -> str:
   return f"{short_benchmark}{short_id}"
 
 
-def _run_workload(
+@task
+def run_workload(
```
RissyRan, line 44: Thanks for this change! You tried it on both local and composer env right? I previously tried to remove `kubernetes` dependency, but met one issue (sorry I did not recall what it was). So would like to double check. 

Could you check if any dependency from https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/.github/requirements.txt could be removed? If so, could you also help remove that from terraform template (if needed)? 

i.e.
```
apache-airflow-providers-docker
...
```

xlml/utils/xpk.py
RissyRan: Thanks! Do you know why this task is not scheduled to run? https://screenshot.googleplex.com/9UPHdZx8oAqwbkz

xlml/utils/xpk.py
jonb377: Sure! I can remove that one, do you know if `kubernetes` is still required?

docs/contributing.md
```diff
@@ -65,6 +65,20 @@ scripts/local-airflow.sh path/to/dag_file.py
 
 Comment out any test cases in the DAG that you do not want to run, or create a temporary DAG file to avoid running all tests.
 
+
+##### XPK-based tests
+
+When running Airflow locally, XPK requires kubectl to support GKE-based authentication. Install the required component using
+
+```
+gcloud components install gke-gcloud-auth-plugin
```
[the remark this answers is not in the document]
jonb377: It's not a pip package, so unfortunately it needs a specific CLI installation.

xlml/utils/xpk.py
jonb377: Ah it's a bit outdated, I hadn't added the GCS support here. I'll do another run shortly.

xlml/utils/xpk.py
RissyRan: Good question! Could you double check with the IDE and Pylint logs (if you run code-style.sh). If no import is in grey, I think we should be fine.

jonb377: I see, that makes sense. I don't see any usage of KubernetesPodOperator or task.kubernetes, so I'll remove it and see if the DagCheck passes. At any rate, there's no harm in keeping it since it's part of the Composer env.

jonb377: Looks like DAG check is green!

RissyRan (approved):

RissyRan (commented):
Could you also remove L146 in TF temp? It's a comment.

xlml/utils/xpk.py
will-cromar: We still need the `kubernetes` package for the GKE/GPU tests

jonb377: That's just `kubernetes` and not the Airflow package though, right? This is the one to be removed: `apache-airflow-providers-cncf-kubernetes`

xlml/apis/task.py
```diff
@@ -374,6 +377,25 @@ def run_model(self) -> DAGNode:
...
+      wait_for_workload_completion = xpk.wait_for_workload_completion.override(
+          timeout=self.task_test_config.time_out_in_min * 60,
+      )(
+          workload_id=workload_id,
+          project_id=self.task_gcp_config.project_name,
+          region=self.task_gcp_config.zone[:-2],
+          cluster_name=self.task_test_config.cluster_name,
+      )
+
+      (workload_id, gcs_path) >> launch_workload >> wait_for_workload_completion
```
will-cromar, line 393: You don't need to add an explicit control dependency for `workload_id` and `gcs_path` because `launch_workload` already consumes the results.

xlml/utils/xpk.py
```diff
@@ -37,98 +40,122 @@ def generate_workload_id(benchmark_id: str) -> str:
...
+) -> bool:
+  """Check if the workload has started."""
+  core_api = _get_core_api_client(project_id, region, cluster_name)
+  pods = _list_workload_pods(core_api, workload_id)
+  print(f"Found {len(pods.items)} pods for workload {workload_id}")
+  return len(pods.items) > 0
+
+
+@task.sensor(poke_interval=60, timeout=600, mode="reschedule")
+def wait_for_workload_completion(
```
will-cromar, line 121: This looks similar to `stream_logs` in `gke.py`. Can these be combined? https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/6465233478e1bee0a24a1c9019c623b05220e11a/xlml/utils/gke.py#L71

My implementation is flawed, so I'd like to piggy-back on any improvements you've made :smile:

xlml/utils/xpk.py
```diff
@@ -37,98 +40,122 @@ def generate_workload_id(benchmark_id: str) -> str:
...
     workload_id: str,
+    gcs_path: str,
     docker_image: str,
     accelerator_type: str,
     run_cmds: str,
     num_slices: int = 1,
 ):
   """Run workload through xpk tool."""
-  import subprocess
+  from subprocess import run, STDOUT, PIPE
```
will-cromar, line 58: Do these have to be imported inside `run_workload`? This is a plain `@task`, so I expect global imports to work.

xlml/utils/xpk.py
```diff
@@ -37,98 +40,122 @@ def generate_workload_id(benchmark_id: str) -> str:
...
       (
-          "python3 xpk.py workload create"
+          "python ${TMPDIR}/xpk/xpk.py workload create"
           f" --cluster={cluster_name} --workload={workload_id}"
           f" --command='{run_cmds}' --device-type={accelerator_type}"
           f" --num-slices={num_slices} --docker-image={docker_image}"
+          f" --project={cluster_project} --zone={zone}"
       ),
+      "rc=$?",
+      "rm -rf $TMPDIR",
```
will-cromar, line 76: Can you use a standard context manager here to manage the temp directory? https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

xlml/utils/xpk.py
```diff
@@ -37,98 +40,122 @@ def generate_workload_id(benchmark_id: str) -> str:
...
+      "exit $rc",
   )
 
-  subprocess.run(["bash", "-c", ";".join(cmds)], check=True)
+  res = run(["bash", "-c", ";".join(cmds)], stderr=STDOUT, stdout=PIPE)
```
will-cromar, line 80: `SubprocessHook` should work now since you're not in a container: https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/hooks/subprocess/index.html#airflow.hooks.subprocess.SubprocessHook

You can pass in the env as a dict instead of tacking on `export`s, and the logs get plumbed automatically.

xlml/utils/xpk.py
will-cromar: Yeah, removing the kubernetes airflow plugin is fine.

xlml/utils/xpk.py
jonb377: Ah great point! This is Python we're working in.

xlml/utils/xpk.py
jonb377: I thought about this, but for XPK the workloads can be long-running (up to 5 hours for convergence tests), so I thought to keep it as a sensor to avoid blocking an Airflow worker on the training job.

We're also generally looking at the workloads after execution completes, so I think the cons of occupying the Airflow workers outweighs the benefit, for XPK at least.

Let me know what you think.

xlml/utils/xpk.py
jonb377: They could be global, but I believe @RissyRan mentioned it's better to import locally for dag parsing time?

xlml/apis/task.py
jonb377: I discussed with Ran, since we already have the dependency chain we wanted to keep this part explicit as well. I don't mind either way.

xlml/utils/xpk.py
will-cromar: That makes sense for big libraries like TF, but these two imports are going to be more-or-less instant.

RissyRan: I think we could have some best practices during refactors. Sometimes it's hard to tell if this dependency is big or small.

xlml/apis/task.py
RissyRan: Yeah, either remove this line or keep it correct for less confusion.

PR #210, opened 2024-03-20
Support of multiple zone selection for GCE.

zpcore:
# Description

1. Support GCE GPU configs to take multiples zones as input, e.g.,
```
  # Running on A100 GPU
  config.get_torchbench_gpu_config(
      machine_type=resource.MachineVersion.A2_HIGHGPU_1G,
      image_project=resource.ImageProject.DEEP_LEARNING_PLATFORM_RELEASE,
      image_family=resource.ImageFamily.COMMON_CU121_DEBIAN_11,
      accelerator_type=resource.GpuVersion.A100,
      count=1,
      gpu_zone=[resource.Zone.US_CENTRAL1_F, resource.Zone.US_CENTRAL1_A],    <------ accept both list or single region.
      model_name=model,
      time_out_in_min=1600,
      extraFlags=" ".join(torchbench_extra_flags),
  ).run()
```

Note: gpu_zone must change to use List after the PR.

2. Add the feature to switch to the next zone if the reservation of resource in current zone failed.

# Tests

[single zone](https://402f31b802dc42918d8712d630ea643a-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla-torchbench/grid?dag_run_id=manual__2024-03-25T16%3A44%3A41.143610%2B00%3A00&tab=graph).
[multiple zone(example that the first zone fail)](https://402f31b802dc42918d8712d630ea643a-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla-torchbench/grid?dag_run_id=manual__2024-03-25T17%3A25%3A54.033900%2B00%3A00&tab=graph)

# Checklist


dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
```diff
@@ -272,16 +272,20 @@ def get_torchbench_gpu_config(
     image_family: resource.ImageFamily,
     accelerator_type: resource.GpuVersion,
     count: int,
-    gpu_zone: resource.Zone,
+    gpu_zone: Union[resource.Zone, List[resource.Zone]],
```
RissyRan, line 275: Do you think it may be better if we just define it as `List[resource.Zone]]` even for a single zone? i.e. [`us-central2-b`]? so we don't need a if/else statement.

xlml/apis/gcp_config.py
```diff
@@ -32,7 +33,7 @@ class GCPConfig:
   """
 
   project_name: str
-  zone: str
+  zone: Union[str, List[str]]
```
RissyRan, line 36: Same comment above.

xlml/utils/gpu.py
```diff
@@ -109,12 +113,30 @@ def generate_gpu_name() -> str:
   return f"gpu-{str(uuid.uuid4())}"
 
 
+@task(provide_context=True)
+def select_zone(gcp: gcp_config.GCPConfig) -> airflow.XComArg:
+  context = get_current_context()
+  try_count = context["task_instance"].try_number - 1
```
RissyRan, line 119: Do we need to specify try_number based on zone? i.e. we have 4 zones to try, however, the default try_number is 2.

RissyRan (commented):
A high level question, what happens if:

1) enable a list of zones
2) create_resource task group still has create_resource_request, wait_for_creation and get_ip_address
3) when wait_for_creation fails, retry with create_resource_request but based on calculate number we select the zone. i.e. no retry: 1st element; 1st retry: 2nd element, etc.
4) so we won't need to introduce more tasks

Is this feasible? HDYT?

xlml/utils/gpu.py
zpcore: My understanding is that the `context["task_instance"].try_number` will be automatically  +1 updated everytime we  retry the task.

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
zpcore: Updated to list format.

zpcore:
> A high level question, what happens if:
> 
> 1. enable a list of zones
> 2. create_resource task group still has create_resource_request, wait_for_creation and get_ip_address
> 3. when wait_for_creation fails, retry with create_resource_request but based on calculate number we select the zone. i.e. no retry: 1st element; 1st retry: 2nd element, etc.
> 4. so we won't need to introduce more tasks
> 
> Is this feasible? HDYT?

Hmm, I was thinking about this before. The issue is that `zone` is not only used in provision but also in later tasks like `clean_up`. If we remove `gpu.select_zone` task we need to let task `gpu.create_resource` return both ip_address and the zone. However, one task only allow to return a single `airflow.XComArg`.  We can't let it return a tuple like [str, str] or `Tuple[airflow.XComArg, airflow.XComArg]`.

xlml/utils/gpu.py
```diff
@@ -285,8 +311,71 @@ def wait_for_resource_creation(operation_name: airflow.XComArg):
...
+
+    Args:
+      context: Airflow context that provides task info.
+      session: Ariflow session info for clearing tasks.
+    """
+    current_task = context["task"]
+    visited = set()
+    queue = [current_task]
+    # Graph search for all upstream tasks.
+    while queue:
```
will-cromar, line 329: To be honest, I'm really not comfortable with this sort of graph introspection.  One of the benefits of airflow IMO is that it's very easy to look at the graph view and see what the control flow would be. Tasks modifying each others' states starts to destroy this expectation. It's going to get messy and hard to reason about very quickly. Let's try to find another solution.

xlml/apis/gcp_config.py
```diff
@@ -32,7 +33,7 @@ class GCPConfig:
   """
 
   project_name: str
-  zone: str
+  zone: List[str]
```
will-cromar, line 36: I'm kind of surprised that some checks didn't fail with this change, since it touches basically every single test in this repository, which have not been updated to use a list. Thanks [duck typing](https://en.wikipedia.org/wiki/Duck_typing).

Tying a retry policy to the `GCPConfig` doesn't seem right to me. It seems like it's mixing levels of abstraction in a way that doesn't seem healthy.

Retry policies that use different zones seem like a headache in general: how do you handle regionalized depedencies like networks or data? But, if we absolutely have to go down this road, I think it makes more sense to leave `GCPConfig` as-is and instead have your `Task` take a list of `GCPConfig`s if it supports retries. Retries may need to take place in multiple projects, or across multiple GKE clusters (which will soon be captured as an extension of GCPConfig).

will-cromar:
You can have multiple returns within taskflow: https://airflow.apache.org/docs/apache-airflow/stable/tutorial/taskflow.html#multiple-outputs-inference

RissyRan (commented):
Also discussed offline.

xlml/utils/gpu.py
```diff
@@ -250,9 +273,12 @@ def create_resource_request(
...
 
   @task.sensor(
-      poke_interval=60, timeout=timeout.total_seconds(), mode="reschedule"
+      poke_interval=60,
+      timeout=min(60 * 4, timeout.total_seconds()),
+      mode="reschedule",
   )
-  def wait_for_resource_creation(operation_name: airflow.XComArg):
+  def wait_for_resource_creation(operation_name: str, zone: str):
+    logging.info(f"Creating the {operation_name} operation_name in {zone}...2")
```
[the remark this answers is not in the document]
zpcore: Oh, typo

xlml/utils/gpu.py
zpcore: Right, I will switch to GKE instead of modifying each others task status.

PR #211, opened 2024-03-21
Add MaxText stable and nightly profiling test

SurbhiJainUSC:
# Description

Add MaxText stable and nightly profiling tests.

# Tests

Ran the tests locally: http://shortn/_9b98QFU9fh

# Checklist

### DIFF
--- a/dags/multipod/maxtext_profiling.py
+++ b/dags/multipod/maxtext_profiling.py
@@ -52,7 +52,7 @@ with models.DAG(
       "python3 MaxText/tests/profiler_test.py",
       "pip3 uninstall -y tensorboard_plugin_profile",
       "pip3 install tbp-nightly",
-      "python3 MaxText/tests/profiler_test.py"
+      "python3 MaxText/tests/profiler_test.py",
   )
 
   for mode, image in docker_images:
--- a/dags/multipod/maxtext_profiling.py
+++ b/dags/multipod/maxtext_profiling.py
@@ -41,28 +41,28 @@ with models.DAG(
   ]
 
   profiling_cmds = (
-    "export RUN_NAME='profiling_tests_$(date +%Y-%m-%d-%H-%M-%S)'",
-    "python3 MaxText/train.py MaxText/configs/base.yml"
-    f" run_name=$RUN_NAME base_output_directory={base_output_directory}"
-    f" dataset_path={dataset_path} enable_profiler=true steps=20",
-    f"gsutil cp -R {base_output_directory}/$RUN_NAME/tensorboard .",
-    "pip3 uninstall -y tbp-nightly",
-    "pip3 uninstall -y tensorboard_plugin_profile",
-    "pip3 install tensorboard_plugin_profile",
-    "python3 MaxText/tests/profiler_test.py",
-    "pip3 uninstall -y tensorboard_plugin_profile",
-    "pip3 install tbp-nightly",
-    "python3 MaxText/tests/profiler_test.py"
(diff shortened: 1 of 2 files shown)


dags/multipod/maxtext_profiling.py
```diff
@@ -0,0 +1,68 @@
...
+) as dag:
+  base_output_directory = gcs_bucket.XLML_OUTPUT_DIR
+  dataset_path = gcs_bucket.MAXTEXT_DIR
+  docker_images = [
+      (SetupMode.STABLE, DockerImage.MAXTEXT_JAX_STABLE),
+      (SetupMode.NIGHTLY, DockerImage.MAXTEXT_JAX_NIGHTLY),
+  ]
+
+  profiling_cmds = (
+      "export RUN_NAME='profiling_tests_$(date +%Y-%m-%d-%H-%M-%S)'",
```
jonb377, line 44: Can you double check that this is being formatted properly in the command? With `'`, the `$(date ...)` should be treated as a string literal instead of being evaluated and substituted. If it looks fine, we might have an issue with how we're escaping the bash commands in the infra.

Also, this could collide for the nightly and stable tests if they launch at the same time. Maybe move this inside the for loop and append the test mode, e.g. `f"export RUN_NAME=profiling_tests_$(date ...)_{mode.value}"`

jonb377 (commented):
Thanks Surbhi!

dags/multipod/maxtext_profiling.py
SurbhiJainUSC: Yes, I could see the run_name being propagated properly: http://shortn/_JbijteLw8L

Sounds good, I will append the test mode.

jonb377: Ah, thanks for confirming! This is an issue then. Could you try removing the `'` for now? I'll work on a fix in the infra, don't want to accidentally break your tests in the future.

SurbhiJainUSC: It works without `'`, so the command in my test will look like: `f"export RUN_NAME=profiling_tests_{mode.value}_$(date +%Y-%m-%d-%H-%M-%S)"`

dags/multipod/maxtext_profiling.py
```diff
@@ -0,0 +1,68 @@
...
+
+with models.DAG(
+    dag_id="maxtext_profiling",
+    schedule=SCHEDULED_TIME,
+    tags=["multipod_team", "maxtext", "stable", "nightly"],
+    start_date=datetime.datetime(2024, 3, 1),
+    catchup=False,
+    concurrency=2,
+) as dag:
+  base_output_directory = gcs_bucket.XLML_OUTPUT_DIR
```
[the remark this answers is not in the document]
RissyRan: Thanks Jon! I have a [PR](https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/212) today to unify our flax tests location, and would recommend we use default gcs folder for each test including xpk. HDYT?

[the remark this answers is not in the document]
jonb377: Agreed, let's get XPK onboarded to this. @SurbhiJainUSC could you follow the format here for now, and I'll unify the GCS path in a follow-up: https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/067303ba64a25183ca7b592632c06507a10d1284/xlml/utils/name_format.py#L61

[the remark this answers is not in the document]
SurbhiJainUSC: Fixed the GCS path for test output.

RissyRan (approved):
LGTM! Thanks!

jonb377 (approved):

PR #212, opened 2024-03-21
Generalize gcs location for tests

RissyRan:
# Description

Currently, we can generate a dedicated gcs folder for each test - [source](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/067303ba64a25183ca7b592632c06507a10d1284/xlml/utils/name_format.py#L51), and this task is in provision in GCE track. 

We could have a separate PR to enable this for GKE track (xpk), so:
* we don't need to indicate `use_runtime_generated_gcs_folder` flag & base_output_directory
* all tests results are stored in dedicated folder based on benchmark ID.

Gradually, we want to reduce the gap between XLML functional tests and benchmarking tests. In this PR:
* Fix issue in TensorBoard summary processing
* Consolidate both func and benchmarking outputs into `output` folder
* Update Flax conv tests' GCS folder to use the default one.

# Tests

Upload to airflow and test.

I only tested BERT as both BART and ViT conv tests are failed before post_process step - [link](https://screenshot.googleplex.com/B9LsYZvNsUgC7fZ)

# Checklist

### DIFF
--- a/dags/multipod/configs/gke_config.py
+++ b/dags/multipod/configs/gke_config.py
@@ -108,7 +108,7 @@ def get_gke_maxtext_nightly_config(
   current_date = current_time.strftime("%Y-%m-%d")
   current_datetime = current_time.strftime("%Y-%m-%d-%H-%M-%S")
   base_output_directory = (
-      f"{gcs_bucket.XLML_OUTPUT_DIR}/maxtext/nightly/automated/{current_date}"
+      f"{gcs_bucket.BASE_OUTPUT_DIR}/maxtext/nightly/automated/{current_date}"
   )
   run_name = f"{num_slices}slice-V{tpu_version.value}_{tpu_cores}-maxtext-nightly-{current_datetime}"
 
@@ -173,7 +173,7 @@ def get_gke_gpt3_6b_nightly_config(
   current_date = current_time.strftime("%Y-%m-%d")
   current_datetime = current_time.strftime("%Y-%m-%d-%H-%M-%S")
   base_output_directory = (
-      f"{gcs_bucket.XLML_OUTPUT_DIR}/maxtext/nightly/automated/{current_date}"
+      f"{gcs_bucket.BASE_OUTPUT_DIR}/maxtext/nightly/automated/{current_date}"
   )
   run_name = f"{num_slices}slice-V{tpu_version.value}_{tpu_cores}-gpt3-6b-nightly-{current_datetime}"
 
--- a/dags/multipod/configs/mxla_collective_config.py
+++ b/dags/multipod/configs/mxla_collective_config.py
@@ -50,7 +50,7 @@ def get_mxla_collective_config(
   current_date = current_time.strftime("%Y-%m-%d")
   current_datetime = current_time.strftime("%Y-%m-%d-%H-%M-%S")
 
-  base_output_directory = f"{gcs_bucket.XLML_OUTPUT_DIR}/multipod/mxla/nightly/automated/{current_date}/{num_slices}slice-V{tpu_version.value}_{tpu_cores}-mxla-collective-{bytes_to_transfer}transferBytes-{current_datetime}"
+  base_output_directory = f"{gcs_bucket.BASE_OUTPUT_DIR}/multipod/mxla/nightly/automated/{current_date}/{num_slices}slice-V{tpu_version.value}_{tpu_cores}-mxla-collective-{bytes_to_transfer}transferBytes-{current_datetime}"
 
   test_platform = common.Platform.GCE
   set_up_cmds = common.setup_mxla_collective()
--- a/dags/solutions_team/configs/flax/solutionsteam_flax_latest_supported_config.py
+++ b/dags/solutions_team/configs/flax/solutionsteam_flax_latest_supported_config.py
@@ -21,6 +21,7 @@ from dags import gcs_bucket, test_owner
 from dags.solutions_team.configs.flax import common
 from dags.vm_resource import TpuVersion, Project, RuntimeVersion
 from datetime import datetime
+import os
 
 
 PROJECT_NAME = Project.CLOUD_ML_AUTO_SOLUTIONS.value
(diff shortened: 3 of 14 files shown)


RissyRan:
@jonb377 Does your team want to use this default gcs bucket for each xpk test? we could definitely add this task there.

dags/multipod/configs/mxla_collective_config.py
```diff
@@ -50,7 +50,7 @@ def get_mxla_collective_config(
   current_date = current_time.strftime("%Y-%m-%d")
   current_datetime = current_time.strftime("%Y-%m-%d-%H-%M-%S")
 
-  base_output_directory = f"{gcs_bucket.XLML_OUTPUT_DIR}/multipod/mxla/nightly/automated/{current_date}/{num_slices}slice-V{tpu_version.value}_{tpu_cores}-mxla-collective-{bytes_to_transfer}transferBytes-{current_datetime}"
+  base_output_directory = f"{gcs_bucket.OUTPUT_DIR}/multipod/mxla/nightly/automated/{current_date}/{num_slices}slice-V{tpu_version.value}_{tpu_cores}-mxla-collective-{bytes_to_transfer}transferBytes-{current_datetime}"
```
zpcore, line 53: `current_datetime` will keep updating every time the file is been imported. Is the datetime used as an unique id to prevent the file using the same name as other runs?

dags/solutions_team/configs/flax/solutionsteam_flax_latest_supported_config.py
```diff
@@ -434,11 +433,11 @@ def get_flax_bart_conv_config(
 
   set_up_cmds = get_flax_bart_setup_cmds()
   work_dir = "/tmp/transformers/bart-base-wiki"
-  tf_summary_location = f"{work_dir}/events.out.tfevents.flax-bart.v2"
-  gcs_location = f"{gcs_bucket.XLML_OUTPUT_DIR}/flax/bart/{RUN_DATE}/events.out.tfevents.flax-bart.v2"
+  file_name = "events.out.tfevents.flax-bart.v2"
+  tf_summary_location = f"{work_dir}/{file_name}"
```
zpcore, line 437: I think you can use path.join here and below also.

zpcore (approved):

jonb377:
> @jonb377 Does your team want to use this default gcs bucket for each xpk test? we could definitely add this task there.

Yeah, this looks like a good fit. I see it's exposed to the workload as an environment variable - I guess the issue with keeping it in xcom is string formatting (i.e. the command formatting happens at DAG evaluation time, but the XComArg isn't generated until run time)?

It would be nice to keep it in XCom since xpk doesn't have great environment override support (we'd need to inject an `export` in the workload command), but either way we can definitely reuse this infrastructure.

dags/multipod/configs/mxla_collective_config.py
RissyRan: Yeah, @jonb377 do you know if your team meet any issue for this? We'd like to recommend to have a similar change for GCE track. Also, add this similar feature for xpk (happy to work together on this).

jonb377: This is only safe for the specific use case - the output directory isn't reused across tasks, so whatever value is in the DAG at execution time is what the workload uses. No downstream tasks depend on what value the workload used.

We will be moving to the infra-generated buckets soon.

RissyRan:
> > @jonb377 Does your team want to use this default gcs bucket for each xpk test? we could definitely add this task there.
> 
> Yeah, this looks like a good fit. I see it's exposed to the workload as an environment variable - I guess the issue with keeping it in xcom is string formatting (i.e. the command formatting happens at DAG evaluation time, but the XComArg isn't generated until run time)?
> 
> It would be nice to keep it in XCom since xpk doesn't have great environment override support (we'd need to inject an `export` in the workload command), but either way we can definitely reuse this infrastructure.

Yeah, currently this gcs location is the XCom from provision group, please see [here](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/task.py#L167). We don't need to use ssh env, which is needed for GCE track. Happy to chat more for xpk!

dags/solutions_team/configs/flax/solutionsteam_flax_latest_supported_config.py
RissyRan: Sure, updated.

dags/multipod/configs/mxla_collective_config.py
RissyRan: SG! Thank you!

dags/solutions_team/configs/flax/solutionsteam_flax_latest_supported_config.py
```diff
@@ -176,16 +177,14 @@ def get_flax_vit_conv_config(
...
-  gcs_location = f"{gcs_bucket.XLML_OUTPUT_DIR}/flax/vit/{RUN_DATE}/events.out.tfevents.flax-vit.v2"
+  file_name = "events.out.tfevents.flax-vit.v2"
+  tf_summary_location = f"/tmp/transformers/vit-imagenette/{file_name}"
   extra_run_cmds = (
       (
           "cp /tmp/transformers/vit-imagenette/events.out.tfevents.*"
           f" {tf_summary_location} || exit 0"
       ),
-      f"gsutil cp {tf_summary_location} {gcs_location} || exit 0",
+      f"gsutil cp {tf_summary_location} ${metric_config.SshEnvVars.GCS_OUTPUT.value} || exit 0",
```
will-cromar, line 187: nit: the layers of interpolation here are breaking my brain a bit, since `{` has a specific and similar meaning in both f-strings and in `[ba]sh`.

What do you think of actually making the values of this enum include the `${}`? So your f-string would read `f"gsutil cp {tf_summary_location} {metric_config.SshEnvVars.GCS_OUTPUT.value}"`, which then evaluates to `gsutil cp /tmp/etc ${GCS_OUTPUT}` before being sent to bash?

dags/multipod/configs/gke_config.py
```diff
@@ -108,7 +108,7 @@ def get_gke_maxtext_nightly_config(
   current_date = current_time.strftime("%Y-%m-%d")
   current_datetime = current_time.strftime("%Y-%m-%d-%H-%M-%S")
   base_output_directory = (
-      f"{gcs_bucket.XLML_OUTPUT_DIR}/maxtext/nightly/automated/{current_date}"
+      f"{gcs_bucket.OUTPUT_DIR}/maxtext/nightly/automated/{current_date}"
```
will-cromar, line 111: nit: can we clarify the difference between `OUTPUT_DIR` and `GCS_OUTPUT`? I think these are similar enough to cause confusion in the future. What do you think of calling `OUTPUT_DIR` something like `GCS_OUTPUT_BASE_DIR`?

will-cromar (approved):
Thanks!

Overall LGTM with some minor nits about naming and formatting.

dags/multipod/configs/gke_config.py
RissyRan: SG! I will update that.

jonb377 (approved):
LGTM

dags/multipod/configs/gke_config.py
RissyRan: I will update it to `BASE_OUTPUT_DIR` as it comes from `gcs_bucket` file, which indicates it belongs to gcs already.

dags/solutions_team/configs/flax/solutionsteam_flax_latest_supported_config.py
RissyRan: Updated, and re-tested: https://screenshot.googleplex.com/84uU3JL6pR6dtFo

sshahrokhi (approved):

PR #220, opened 2024-03-25
Migrate multislice PyTorch tests to Airflow

jonb377:
# Description

Migrate multislice PyTorch tests to Airflow.

Also, add support for generated GCS paths in XPK. Unfortunately, task.docker and task.kubernetes execute in an isolated environment without any dependencies available, so we can't directly reference the `metric_config` module to retrieve the value. This requires hard-coding the GCS_OUTPUT environment variable's name into the XPK workload creation command.

# Tests

Ran the tests locally (metrics failures are expected with local airflow): https://screenshot.googleplex.com/BGUoFgA7NBGZ9AA

# Checklist

### DIFF
--- a/dags/multipod/configs/pytorch_config.py
+++ b/dags/multipod/configs/pytorch_config.py
@@ -1,0 +1,56 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dags.vm_resource import TpuVersion, Zone, DockerImage, ClusterName
+from dags.multipod.configs import gke_config
+from xlml.apis import task
+from typing import List
+
+
+# TODO(jonbolin): Refactor this to cluster definition
+CLUSTER_CONFIG = {
+    ClusterName.V4_8_MULTISLICE_CLUSTER: {
+        'tpu_version': TpuVersion.V4,
+        'tpu_cores': 8,
+        'tpu_zone': Zone.US_CENTRAL2_B.value,
+    },
+    ClusterName.V4_16_MULTISLICE_CLUSTER: {
+        'tpu_version': TpuVersion.V4,
+        'tpu_cores': 16,
+        'tpu_zone': Zone.US_CENTRAL2_B.value,
+    },
+}
+
+
+def get_nightly_pytorch_config(
+    test_name: str,
+    test_owner: str,
+    run_commands: List[str],
+    cluster: ClusterName,
+    num_slices: int,
+) -> task.XpkTask:
+  cmds = (
+      'git clone https://github.com/pytorch/xla /pytorch/xla',
+      *run_commands,
+  )
+  return gke_config.get_gke_config(
+      cluster_name=cluster.value,
+      test_name=test_name,
+      run_model_cmds=cmds,
+      num_slices=num_slices,
+      docker_image=DockerImage.PYTORCH_NIGHTLY.value,
+      test_owner=test_owner,
+      time_out_in_min=60,
+      **CLUSTER_CONFIG[cluster],
+  )
--- a/dags/multipod/pytorch.py
+++ b/dags/multipod/pytorch.py
@@ -1,0 +1,69 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+A DAG to run PyTorch multislice tests
+"""
+import datetime
+from airflow import models
+from dags import composer_env, test_owner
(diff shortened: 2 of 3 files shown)


xlml/utils/xpk.py
```diff
@@ -52,6 +53,12 @@ def _run_workload(
   """Run workload through xpk tool."""
   import subprocess
 
+  # Set the GCS path environment variable in the run_cmds
+  # TODO(jonbolin): Can't access `metric_config` from within
+  # task.(docker|kubernetes), so hardcode metric_config.SshEnvVars.GCS_OUTPUT
+  # here
+  run_cmds = f"export GCS_OUTPUT={gcs_path}; {run_cmds}"
```
jonb377, line 60: cc @RissyRan @will-cromar in case there's a better approach. `gcs_path` will be an XComArg, so we can't format outside of the `@task.docker`, and `metric_config` is inaccessible from within `@task.docker`.

RissyRan (commented):
Thanks Jon!

Just 2 high level questions:

1) what's the error for post_processing for local runs? just curious
2) do we have plans to migrate maxtext tests to use this default folder for each execution in coming PRs? currently, I think tests are using customized gcs folders
3) could you help remove `_` underscore for run_workload?

jonb377:
@RissyRan 

For 1, it's the metrics processing task. There are two errors: one is that it tries to query the Composer environment, and the other is that devboxes don't have write access to the development BigQuery dataset. We could just skip this altogether on local runs, but I hesitate to pull the trigger on that since it would be misleading for a breaking change to metrics processing to pass locally (due to being skipped) but fail in prod.
On 2, I'll work with the test owners on that. Some of the tests have a specific format, since their output is consumed by other teams.
On 3, it's unfortunately due to the `task.docker` and `task.kubernetes` optionality. I think we should migrate this to a raw `@task` to address that, but we need `gcloud` support on the Airflow workers to run XPK. This is a TODO on me right now.

xlml/apis/task.py
```diff
@@ -374,6 +374,9 @@ def run_model(self) -> DAGNode:
     """
     with TaskGroup(group_id="run_model") as group:
       workload_id = xpk.generate_workload_id(self.task_test_config.benchmark_id)
+      gcs_path = name_format.generate_gcs_folder_location(
```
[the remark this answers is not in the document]
jonb377: TaskFlow will implicitly add the dependencies, so I left it out of L400. Let me know if I should make it explicit.

For the shared GCS bucket use case, how would we configure the multiple workloads in one XPK task? Would we need a `List[task_test_config]` and `List[gcp_config]` in `XpkTask` definition?

If we wan to keep the two as distinct XpkTask with the same GCS path, we could make the GCS path an optional argument to `run`, and if it's missing then do the generation via `name_format.generate_gcs_folder_location`. So something like:

```python
def run(gcs_path: XComArg[str]):
   ...
   gcs_path = gcs_path or name_format.generate_gcs_folder_location(...)
```

The in cases where we need to link multiple workloads to the same bucket, do something like:

```python
shared_gcs_path = name_format.generate_gcs_folder_location('shared-task-id')
# Task to convert chkpt
get_gke_config(...).run(gcs_path=shared_gcs_path)
# Task to run model
get_gke_config(...).run(gcs_path=shared_gcs_path)
```

RissyRan:
> @RissyRan
> 
> For 1, it's the metrics processing task. There are two errors: one is that it tries to query the Composer environment, and the other is that devboxes don't have write access to the development BigQuery dataset. We could just skip this altogether on local runs, but I hesitate to pull the trigger on that since it would be misleading for a breaking change to metrics processing to pass locally (due to being skipped) but fail in prod. On 2, I'll work with the test owners on that. Some of the tests have a specific format, since their output is consumed by other teams. On 3, it's unfortunately due to the `task.docker` and `task.kubernetes` optionality. I think we should migrate this to a raw `@task` to address that, but we need `gcloud` support on the Airflow workers to run XPK. This is a TODO on me right now.

I see. Thanks! For 1), I will have a try with local testing env this week and see if we could work around.

xlml/apis/task.py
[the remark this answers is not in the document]
RissyRan:
```
TaskFlow will implicitly add the dependencies, so I left it out of L400. Let me know if I should make it explicit.
```

Let's update that as well for less confusion for readers like me (since we already have L400 there). We could specify something like (workload_id, generate_gcs_folder_location) >> run_workload >> wait_for_workload_completion

Right, the proposal makes sense. I am also thinking to try this `xcom_push` and `xcom_pull` in tasks. For instance, inside of this [method](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/4a020882beb90c1df8b5595e55a3fb32f429bb8b/xlml/utils/name_format.py#L51),  it queries key first:
* if has value: directly return;
* if no value: generate and return.

If this works, we don't need `shared_gcs_path = name_format.generate_gcs_folder_location('shared-task-id')` and `def run(gcs_path: Optional[XComArg[str]] = None)`.

This is xcom value from `generate_gcs_folder_location` task: https://screenshot.googleplex.com/PoFqg73CUNNHgZr. It's already there.

xlml/utils/xpk.py
jonb377: I think the solution here is to move this to a BashOperator and make XPK a dependency for the Composer env: https://cloud.google.com/composer/docs/composer-2/write-dags#bash-operator

I'll take this in a separate PR.

RissyRan (approved):

xlml/utils/xpk.py
jonb377: Ah that makes sense, I guess xpk is still under rapid development so PyPi can fall behind. I'll stick with `git clone` for now. Will address in this PR: https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/208

jonb377: @RissyRan FYI I'm going to move the GCS support into https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/208 and rebase this PR once that's landed.

RissyRan:
Back to our previous conversation about gcs_location implementation for multiple tasks. I noticed when querying gcs_location in 2nd test, we need to know benchmark id for 1st test (we cannot make it same as Airflow will throw name conflict error). One example [here](https://screenshot.googleplex.com/3A2gTPswWdsKwtS).

Currently we don't have a use case for GCE to re-use location, so I agree that it's better to move this logic to run() inside of xpk task.

jonb377:
Rebased and ran the tests locally: https://screenshot.googleplex.com/AzdffiP8mKBX7j7

PR #235, opened 2024-04-02
Nina/fix gpu xpk cmd

NinaCai:
# Description

Fix Maxtext GPU end-to-end testing run commands. Use single node for testings.

# Tests

Test in airflow.

One successful run: https://540cef12d3da42ce97b48a97401b6c83-dot-us-central1.composer.googleusercontent.com/dags/maxtext_end_to_end/grid

# Checklist


dags/multipod/configs/gke_config.py
```diff
@@ -163,7 +163,7 @@ def get_maxtext_end_to_end_gpu_gke_test_config(
       zone=gpu_zone,
       dataset_name=metric_config.DatasetOption.XLML_DATASET,
   )
-  run_model_cmds = (f"cd /tmp/maxtext && bash end_to_end/{test_script}.sh",)
+  run_model_cmds = (f"bash end_to_end/{test_script}.sh",)
```
jonb377, line 166: Just for context, how do the GPU XPK tests run? 

* I see in [xpk.py](https://github.com/google/xpk/blob/main/xpk.py#L259-L260) the command is formatted into the environment, how does it get executed? 
* Is the working directory already `/tmp/maxtext` or is the install directory different for GPU images?

dags/multipod/maxtext_end_to_end.py
```diff
@@ -69,7 +69,6 @@
           time_out_in_min=300,
           test_name=f"{test_name_prefix}-stable-{test_script}",
           test_script=test_script,
-          num_slices=2,
```
jonb377, line 72: Should we remove `num_slices` as an option in the GPU test config?

jonb377 (approved):
LGTM, thanks Nina!

dags/multipod/maxtext_end_to_end.py
NinaCai: We could actually use multiple nodes for gpu testing, but the argument is `--num_nodes` instead of `--num_slices` in the xpk command.

dags/multipod/configs/gke_config.py
NinaCai: 1. It will run [gpu_multi_process_run.sh](https://github.com/google/xpk/blob/8ccfda556ce98462ec67c053293c0846188ea901/xpk.py#L266), and run [commands](https://github.com/google/maxtext/blob/33dd61ef96d11277cac921b031d52de720b54186/gpu_multi_process_run.sh#L148) in multiple processes. 
2. Gpu image is the same, the working directly is already `/tmp/maxtext`

xlml/utils/xpk.py
```diff
@@ -60,14 +61,18 @@ def run_workload(
   """Run workload through xpk tool."""
 
   with tempfile.TemporaryDirectory() as tmpdir:
+    if accelerator_type == GpuVersion.XPK_H100.value:
+      multi_keyword = "num-nodes"
+    else:
+      multi_keyword = "num-slices"
```
jonb377, line 67: @Obliviour Do you think we could consolidate the two flags upstream in xpk? Similar to how we use `device-type` for both TPU and GPU.

xlml/utils/xpk.py
```diff
@@ -60,14 +61,18 @@ def run_workload(
   """Run workload through xpk tool."""
 
   with tempfile.TemporaryDirectory() as tmpdir:
+    if accelerator_type == GpuVersion.XPK_H100.value:
```
jonb377, line 64: Is H100 the only supported GPU type?

jonb377 (approved):
The new approach LGTM, my only concern is on the maintainability in case of other GPU versions besides H100.

Let's not block on it for now, we can see about moving the logic upstream to xpk going forward.

xlml/utils/xpk.py
NinaCai: We might have A3+ too, which is also h100, but with a different keyword than `h100-80gb-8`.

RissyRan (approved):

PR #243, opened 2024-04-08
separate gpu and tpu scripts for end-to-end testing

NinaCai:
# Description

Based on maxtext PR: https://github.com/google/maxtext/pull/578. We will separate tpu and gpu scripts for end-to-end testing.

# Tests

https://540cef12d3da42ce97b48a97401b6c83-dot-us-central1.composer.googleusercontent.com/dags/maxtext_end_to_end/grid?dag_run_id=manual__2024-04-11T02%3A04%3A51.094374%2B00%3A00&tab=graph

# Checklist

### DIFF
--- a/dags/multipod/maxtext_end_to_end.py
+++ b/dags/multipod/maxtext_end_to_end.py
@@ -34,14 +34,18 @@ with models.DAG(
     catchup=False,
 ) as dag:
   test_name_prefix = "maxtext"
-  test_models = {
-      "llama2-7b": "test_llama2_7b",
-      "mistral": "test_mistral",
-      "gemma-2b": "gemma/2b/test_gemma",
-      "gpt3": "test_gpt3",
+  test_models_tpu = {
+      "llama2-7b": "tpu/test_llama2_7b",
+      "mistral": "tpu/test_mistral",
+      "gemma-2b": "tpu/gemma/2b/test_gemma",
+      "gpt3": "tpu/test_gpt3",
   }
 
-  for model, test_script in test_models.items():
+  test_models_gpu = {
+      "llama2-7b": "gpu/test_llama2_7b",
+  }
+
+  for model, test_script in test_models_tpu.items():
     stable_tpu = gke_config.get_gke_config(
         tpu_version=TpuVersion.V4,
         tpu_cores=8,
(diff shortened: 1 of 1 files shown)


dags/multipod/maxtext_end_to_end.py
```diff
@@ -34,14 +34,18 @@
...
+  test_models_tpu = {
+      "llama2-7b": "tpu/test_llama2_7b",
+      "mistral": "tpu/test_mistral",
+      "gemma-2b": "tpu/gemma/2b/test_gemma",
+      "gpt3": "tpu/test_gpt3",
   }
 
-  for model, test_script in test_models.items():
+  test_models_gpu = {
+      "llama2-7b": "gpu/test_llama2_7b",
```
jonb377, line 45: Is there a different run_name between the TPU and GPU tests? Since there's no longer a dependency between the two hardware types, the tests may execute concurrently. We've seen issues when the run_name collides in the past (primarily on checkpointing-related tests).

NinaCai: > Is there a different run_name between the TPU and GPU tests? Since there's no longer a dependency between the two hardware types, the tests may execute concurrently. We've seen issues when the run_name collides in the past (primarily on checkpointing-related tests).

1. Are you referring to the name shown in the Airflow? They have hardware type as suffix.
2. Could you elaborate more on the collision issue? Is it because they save to the save checkpoint path? I changed the path to include `gpu` so that the path is unique. But if it is something else, I would love to know more about it.

jonb377: Ah I mean in MaxText: https://github.com/google/maxtext/blob/main/end_to_end/test_llama2_7b.sh#L43

The issue is that `idx` is created with minute granularity, so if two tests launch in the same minute, they'll have the same run_name and target the same GCS bucket. This is currently avoided by having the sequential chaining between tests of the same script (tpu_stable >> tpu_nightly >> gpu_stable >> gpu_nightly).

If you can add e.g. `gpu` to the run_name for GPU tests, it won't be an issue.

NinaCai: Got it, I change the saved path, so that there won't be any overwritten. 

> Ah I mean in MaxText: https://github.com/google/maxtext/blob/main/end_to_end/test_llama2_7b.sh#L43
> 
> The issue is that `idx` is created with minute granularity, so if two tests launch in the same minute, they'll have the same run_name and target the same GCS bucket. This is currently avoided by having the sequential chaining between tests of the same script (tpu_stable >> tpu_nightly >> gpu_stable >> gpu_nightly).
> 
> If you can add e.g. `gpu` to the run_name for GPU tests, it won't be an issue.

Got it, I change the ckpt path for gpu so that there won't be any overwritten.

jonb377 (approved):
LGTM!

PR #253, opened 2024-04-17
Add terraform configs for a quick internal cloud composer deployment

mbzomowski:
# Description

Added a script for installing Terraform and Terraform configs for quickly spinning up a cloud composer 2 env for internal testing.

# Tests

Manually tested

# Checklist


internal_deployment/composer_env.tf
```diff
@@ -0,0 +1,58 @@
+resource "google_composer_environment" "cloud_ml_auto_solutions_env" {
+  provider = google-beta
+  name     = var.env_name
+  region = var.region
+
+  config {
```
RissyRan, line 6: Instead of having a separate copy for those configs ([source](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/deployment/cloud_composer_template.tf#L131)), could we move it to a common file, and reuse for both developer and prod envs?

RissyRan (commented):
Thanks Mark! This will be really helpful!

One comment at high level. Can we still keep those into `development` folder, but with proper prefix, i.e. `developer_composer_env` etc. Or, have a subfolder like `development/developer`. I feel those are not internal specific, external folks can also use this template.

internal_deployment/composer_env.tf
mbzomowski: Sounds good - I moved the cloud composer resource into its own module, and am now referencing that in both ./deployment and ./deployment/deployment_composer_env

mbzomowski:
Tested both ./deployment and ./deployment/deployment_composer_env with `terraform init; terraform plan` to ensure there weren't any configuration issues and everything still works as expected.

will-cromar (approved):
Thanks!

internal_deployment/composer_env.tf
RissyRan: SG! Thanks!

RissyRan:
Thanks!

PR #263, opened 2024-04-23
Support GKE for torchbench

zpcore:
# Description

1. Modify GKE to support storing result to GCS and postprocess data;
2. Update torchbench to use GKE;
3. [Minor] Update r2.3 release test dependency.

Currently only support V100. Will add H100, L4 and A100.

# Tests

[Torchbench using GKE on V100](https://60bb71acc6784780b3bbdbae4ffa636f-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla-torchbench/grid?dag_run_id=manual__2024-05-16T23%3A54%3A36.254218%2B00%3A00&task_id=torchbench_BERT_pytorch-v4-8&tab=graph)
image

[Jsonline config test](https://60bb71acc6784780b3bbdbae4ffa636f-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla-nightly/grid?dag_run_id=manual__2024-05-21T21%3A32%3A18.524828%2B00%3A00&task_id=pt-nightly-resnet50-mp-plugin-fake-v100-x2x1-nvidia-tesla-v100)
image

# Checklist

### DIFF
--- a/dags/vm_resource.py
+++ b/dags/vm_resource.py
@@ -47,6 +47,13 @@ class ImageFamily(enum.Enum):
   COMMON_CU121_DEBIAN_11 = "common-cu121-debian-11"
 
 
+class Region(enum.Enum):
+  """Common GCP regions."""
+
+  # used for GKE
+  US_CENTRAL1 = "us-central1"
+
+
 class Zone(enum.Enum):
   """Common GCP zones."""
 
--- a/xlml/apis/gcp_config.py
+++ b/xlml/apis/gcp_config.py
@@ -17,7 +17,6 @@
 import dataclasses
 
 from dags.vm_resource import Project
-from typing import Optional
 from xlml.apis import metric_config
 
 
--- a/xlml/apis/gcp_config.py
+++ b/xlml/apis/gcp_config.py
@@ -17,6 +17,7 @@
 import dataclasses
 
 from dags.vm_resource import Project
+from typing import Optional
 from xlml.apis import metric_config
 
 
--- a/xlml/apis/task.py
+++ b/xlml/apis/task.py
@@ -18,7 +18,7 @@ import abc
 import dataclasses
 import datetime
 import shlex
-from typing import Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 import airflow
 from airflow.models.taskmixin import DAGNode
 from airflow.utils.task_group import TaskGroup
@@ -635,21 +635,21 @@ class GpuGkeTask(BaseTask):
   """This is a class to set up tasks for GPU on a GKE cluster.
 
   Attributes:
-    image_project: the project that an image belongs to.
-    image_family: the family group that an image belongs to.
+    task_test_config: task configutation.
+    task_gcp_config: gcp related config (e.g., zone, project) for the task.
     cluster_name: Name of the GCP cluster.
     job_create_timeout: Amount of time to wait for all pods to become active.
+    task_metric_config: metric configuration (e.g., result gcs path).
   """
 
-  task_test_config: test_config.JSonnetGpuTest
+  task_test_config: test_config.GpuGkeTest
   task_gcp_config: gcp_config.GCPConfig
   cluster_name: str
   job_create_timeout: datetime.timedelta = datetime.timedelta(minutes=10)
-  # TODO(wcromar): job history metrics
-  # task_metric_config: Optional[metric_config.MetricConfig] = None
+  task_metric_config: Optional[metric_config.MetricConfig] = None
 
   def run(self) -> DAGNode:
-    """Run a test job.
+    """Run a test job and do post data process.
 
     Returns:
       A task group that runs the given test config on a GKE cluster.
@@ -657,15 +657,42 @@ class GpuGkeTask(BaseTask):
     with TaskGroup(
         group_id=self.task_test_config.benchmark_id, prefix_group_id=True
     ) as group:
+      gcs_location = name_format.generate_gcs_folder_location(
+          self.task_test_config.gcs_subfolder,
+          self.task_test_config.benchmark_id,
+      )
+
       job_body = self._get_job_manifest()
-      gke.run_job.override(group_id="run_model")(
+
+      gke_run = gke.run_job.override(group_id="run_model")(
           job_body,
           self.task_gcp_config,
           self.cluster_name,
           self.job_create_timeout,
+          gcs_location,
       )
-
+      post_process = self.post_process(gcs_location)
+      gcs_location >> gke_run >> post_process
     return group
+
+  def post_process(
+      self, result_location: Optional[airflow.XComArg] = None
+  ) -> DAGNode:
+    """Process metrics and metadata, and insert them into BigQuery tables.
+
+    Returns:
+      A DAG node that executes the post process.
+    """
+    with TaskGroup(group_id="post_process") as group:
+      process_id = metric.generate_process_id.override(retries=0)()
+      metric.process_metrics.override(retries=0)(
+          process_id,
+          self.task_test_config,
+          self.task_metric_config,
+          self.task_gcp_config,
+          folder_location=result_location,
+      )
+      return group
 
   def _get_job_manifest(self):
     accelerator = self.task_test_config.accelerator
@@ -673,7 +700,7 @@ class GpuGkeTask(BaseTask):
         "apiVersion": "batch/v1",
         "kind": "Job",
         "metadata": {
-            "generateName": f"{self.task_test_config.benchmark_id}-",
+            "generateName": f"{self.task_test_config.test_name}",
             "labels": {
                 "accelerator": accelerator.name,
                 "benchmarkId": self.task_test_config.benchmark_id,
--- a/xlml/apis/test_config.py
+++ b/xlml/apis/test_config.py
@@ -465,12 +465,8 @@ class JSonnetTpuVmTest(TestConfig[Tpu]):
 
 
 @attrs.define
-class JSonnetGpuTest(TestConfig[Gpu]):
-  """Convert legacy JSonnet test configs into a TestConfig.
-
-  Do not construct directly. Instead, use the `from_*` factory functions which
-  parse pre-compiled JSonnet test configs.
-
+class GpuGkeTest(TestConfig[Gpu]):
+  """
   Attributes:
     test_name: Unique name of this test/model.
     test_command: Command and arguments to execute on the TPU VM.
@@ -481,17 +477,18 @@ class JSonnetGpuTest(TestConfig[Gpu]):
   """
 
   test_name: str
-  test_command: List[str]
   entrypoint_script: List[str]
+  test_command: List[str]
   docker_image: str
   num_hosts: int = 1
+  gcs_subfolder: str = '/tmp/'
 
   @staticmethod
-  def from_pytorch(test_name: str, network='default', subnetwork='default'):
+  def from_pytorch(test_name: str):
     """Parses a compiled legacy JSonnet test config from `tests/pytorch`."""
     test = _load_compiled_jsonnet(test_name)
 
-    return JSonnetGpuTest(
+    return GpuGkeTest(
         test_name=test_name,
         docker_image=f'{test["image"]}:{test["imageTag"]}',
         accelerator=Gpu(
@@ -509,10 +506,8 @@ class JSonnetGpuTest(TestConfig[Gpu]):
 
   @property
   def benchmark_id(self) -> str:
-    return self.test_name
+    return f'{self.test_name}-{self.accelerator.name}'
 
-  # HACK: setup script is used as the entrypoint in the test. Make sure it
-  # invokes the content of `test_script` at the end (e.g. "${@:0}").
   @property
   def setup_script(self) -> str:
     return shlex.join(self.entrypoint_script)
--- a/xlml/utils/gke.py
+++ b/xlml/utils/gke.py
@@ -44,6 +44,7 @@ def run_job(
     gcp: gcp_config.GCPConfig,
     cluster_name: str,
     job_create_timeout: datetime.timedelta,
+    gcs_location: str = '',
 ):
   """Run a batch job directly on a GKE cluster.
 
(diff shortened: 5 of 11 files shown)


xlml/apis/test_config.py
```diff
@@ -480,10 +476,45 @@ class JSonnetGpuTest(TestConfig[Gpu]):
...
 
   test_name: str
-  test_command: List[str]
   entrypoint_script: List[str]
+  test_command: List[str]
   docker_image: str
+  accelerator: Gpu
+  time_out_in_min: str
   num_hosts: int = 1
+  runtime_version: str = 'n/a'
```
will-cromar, line 485: `runtime_version`, `network`, and `subnetwork` should not be necessary for GKE. That's all specified in terraform and baked into the cluster configuration.

xlml/apis/test_config.py
```diff
@@ -480,10 +476,45 @@ class JSonnetGpuTest(TestConfig[Gpu]):
   """
 
   test_name: str
-  test_command: List[str]
   entrypoint_script: List[str]
+  test_command: List[str]
   docker_image: str
+  accelerator: Gpu
+  time_out_in_min: str
```
will-cromar, line 483: timeout should be a `timedelta` so the unit is self-describing

xlml/apis/test_config.py
```diff
@@ -480,10 +476,45 @@ class JSonnetGpuTest(TestConfig[Gpu]):
   """
 
   test_name: str
-  test_command: List[str]
   entrypoint_script: List[str]
+  test_command: List[str]
   docker_image: str
+  accelerator: Gpu
```
will-cromar, line 482: `accelerator`'s type is derived from the parent type's parameter

xlml/apis/test_config.py
```diff
@@ -480,10 +476,45 @@ class JSonnetGpuTest(TestConfig[Gpu]):
...
+  time_out_in_min: str
   num_hosts: int = 1
+  runtime_version: str = 'n/a'
+  network: str = ('default',)
+  subnetwork: str = 'default'
+  gcs_subfolder: str = 'n/a'
```
will-cromar, line 488: If this is optional, make it an `Optional` so the "null" option is explicit. Making it required or having a workable default (like no subfolder) is also fine by me

xlml/apis/task.py
```diff
@@ -637,18 +637,18 @@ class GpuGkeTask(BaseTask):
...
-    image_family: the family group that an image belongs to.
+    task_test_config: task configutation.
+    task_gcp_config: gcp related config (e.g., zone, project) for the task.
     cluster_name: Name of the GCP cluster.
     job_create_timeout: Amount of time to wait for all pods to become active.
+    task_metric_config: metric configuration (e.g., result gcs path).
   """
 
-  task_test_config: test_config.JSonnetGpuTest
+  task_test_config: test_config.PlainGpuTest
```
will-cromar, line 647: I hope you didn't break support for `JSonnetGpuTest`! If `task_test_config` can be either, then this should be a `Union`. Combining the two test types somehow (likely with different constructors) is also fine.

xlml/apis/task.py
```diff
@@ -669,6 +669,52 @@ def run(self) -> DAGNode:
 
     return group
 
+  def fullrun(self) -> DAGNode:
```
will-cromar, line 672: What's the difference here from `run`? Just that metrics are handled properly? If so, please just merge that change into `run`. I told @RissyRan I'd implement those anyway :sweat_smile:

xlml/apis/test_config.py
```diff
@@ -510,13 +541,3 @@ def from_pytorch(test_name: str, network='default', subnetwork='default'):
   @property
   def benchmark_id(self) -> str:
     return self.test_name
-
-  # HACK: setup script is used as the entrypoint in the test. Make sure it
-  # invokes the content of `test_script` at the end (e.g. "${@:0}").
-  @property
```
will-cromar, line 516: Did you mean to delete these?

xlml/utils/gke.py
```diff
@@ -55,10 +56,14 @@ def run_job(
   """
 
   @task
-  def deploy_job():
+  def deploy_job(gcs_location):
+    body['spec']['template']['spec']['containers'][0]['env'].append(
```
will-cromar, line 60: I don't like modifying `body` like this after the fact. `gcs_location` will have to be plumbed in somewhere else since it's an xcom (or depends on one). 

I'll take another look at the code and get back to you. The change makes sense in principle, and there should be a way to plumb dynamic variables to the environment at test time.

xlml/utils/metric.py
```diff
@@ -570,7 +570,13 @@ def get_gce_job_status(
       wait_task = current_dag.get_task(
           task_id=f"{benchmark_id}.provision.create_queued_resource.wait_for_ready_queued_resource"
       )
-    elif isinstance(task_test_config.accelerator, test_config.Gpu):
+    elif isinstance(task_test_config, test_config.PlainGpuTest):
+      # Test using the GKE GPU.
+      wait_task = current_dag.get_task(
+          task_id=f"{benchmark_id}.run_model.stream_logs"
```
will-cromar, line 576: FYI, this does have a different semantic than `wait_for_ready_queued_resource` or `get_ip_address`. Those pertain only to whether the accelerator was created. `stream_log` also streams the running test's logs.

There's a fundamental conceptual difference: some tests create accelerators on demand. For the GKE test, the accelerator is already there and waiting for a job (or finishing another task first).

@RissyRan do you think that could cause any issues with how we interpret failures?

xlml/utils/metric.py
```diff
@@ -570,7 +570,13 @@ def get_gce_job_status(
       wait_task = current_dag.get_task(
           task_id=f"{benchmark_id}.provision.create_queued_resource.wait_for_ready_queued_resource"
       )
-    elif isinstance(task_test_config.accelerator, test_config.Gpu):
+    elif isinstance(task_test_config, test_config.PlainGpuTest):
```
will-cromar, line 573: This should also catch `JSonnetGpuVmTest`

xlml/apis/task.py
```diff
@@ -680,6 +726,7 @@ def _get_job_manifest(self):
                 "accelerator": accelerator.name,
                 "benchmarkId": self.task_test_config.benchmark_id,
             },
+            "result_gcs_location": "",
```
will-cromar, line 729: What's this for?

xlml/apis/task.py
zpcore: `JSonnetGpuTest` is derived from `PlainGpuTest`:
```
@attrs.define(kw_only=True)
class JSonnetGpuTest(PlainGpuTest)
...
```
It's OK here I think?

xlml/apis/test_config.py
zpcore: Hah, I think you left that comment before.

zpcore: Do you want me to delete it?

xlml/utils/metric.py
zpcore: I just updated a separate `get_gke_job_status` to handle this job flow instead of adding too many conditions in existing `get_..._status` functions.

xlml/utils/metric.py
zpcore: Code updated.

xlml/apis/task.py
zpcore: Nice catch, removed!

xlml/apis/task.py
will-cromar: Oooh, I totally missed the inheritance relationship between them. What's the difference between them? Just the static constructor (`from_pytorch`)? In that case, can you just merge these two test configs into one type called `GpuGkeTest`?

xlml/apis/test_config.py
will-cromar: Missed the inheritance. Did you correct the weird hack I noted where the setup_script invokes the test script? If not, can you preserve that comment as a warning?

xlml/apis/test_config.py
will-cromar: I mean the type of this should be `datetime.timedelta`. Calling the variable "timeout` is fine

dags/vm_resource.py
```diff
@@ -50,6 +50,8 @@ class ImageFamily(enum.Enum):
 class Zone(enum.Enum):
   """Common GCP zones."""
 
+  # used for GKE
+  US_CENTRAL1 = "us-central1"
```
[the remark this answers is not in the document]
zpcore: Right, updated

xlml/utils/metric.py
```diff
@@ -545,6 +545,50 @@ def get_gke_job_status(benchmark_id: str) -> bigquery.JobStatus:
   return bigquery.JobStatus.FAILED
 
 
+def get_gke_job_status(
+    task_test_config: test_config.TestConfig[test_config.Accelerator],
+) -> bigquery.JobStatus:
+  """Get job status for the GCE run.
+
+  MISSED - if any failure occurs in initialize & create_queued_resource
+  FAILED - if any failure occurs in setup & run_model (including timeout of
```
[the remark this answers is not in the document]
zpcore: Yes, actually we don't have MISSED status here also. Cleaned them up.

xlml/utils/metric.py
```diff
@@ -545,6 +545,50 @@ def get_gke_job_status(benchmark_id: str) -> bigquery.JobStatus:
...
+  run_model_state = run_model_ti.current_state()
+
+  if run_model_state == TaskState.SUCCESS.value:
+    logging.info(
+        "The run_model state is success, and the job status is success."
+    )
+    return bigquery.JobStatus.SUCCESS
+
+  logging.info("The run_model state is failed, and the job status is failed.")
+  return bigquery.JobStatus.FAILED
```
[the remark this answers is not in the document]
zpcore: Nice catch, I have removed the status.

xlml/apis/test_config.py
zpcore: Do you mean we don't need to mentioned `accelerator: Gpu` here because it is deticed from `TestConfig[Gpu]`?

xlml/apis/task.py
zpcore: Merged into one.

xlml/apis/test_config.py
```diff
@@ -464,12 +464,8 @@ def test_script(self) -> str:
 
 
 @attrs.define
-class JSonnetGpuTest(TestConfig[Gpu]):
-  """Convert legacy JSonnet test configs into a TestConfig.
```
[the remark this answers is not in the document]
zpcore: I removed the class and use `GpuGkeTest` class uniformly.

xlml/apis/test_config.py
zpcore: I cleaned up the xxxGpuTest into `GpuGkeTest`. It looks cleaner now.

xlml/apis/task.py
zpcore: > Yes, we miss post_process for it, but I see the test is green for metric processing?

Because I am calling `fullrun` to handle the metric processing in this case. `fullrun` contains the postprocess.

zpcore: > What's the difference here from `run`? Just that metrics are handled properly? If so, please just merge that change into `run`. I told @RissyRan I'd implement those anyway 😅

If we add postprocess into `run`, then all the nightly test will contains the postprocess step. However, we don't need postprocess for nightly tests. Do we still want to unify those runs?

xlml/utils/gke.py
will-cromar: Discussed offline: please leave a TODO. I'm doing some refactoring and can add the plumbing we need.

xlml/apis/task.py
zpcore: Based on the offline discussion, I unified all gke run to use `run` function.

xlml/apis/gcp_config.py
```diff
@@ -27,13 +28,14 @@ class GCPConfig:
...
+    dataset_name: The option of dataset for metrics. If it is None, we do not
+      postprocess/store results.
     dataset_project: The name of a project that hosts the dataset.
     composer_project: The name of a project that hosts the composer env.
   """
 
   project_name: str
   zone: str
-  dataset_name: metric_config.DatasetOption
+  dataset_name: Optional[metric_config.DatasetOption] = None
```
RissyRan, line 39: No, please don't set it as Optional. The system will always need to send metrics to BigQuery to monitor all tests status.

xlml/utils/metric.py
```diff
@@ -658,6 +700,10 @@ def process_metrics(
   metadata_history_rows_list = [[]]
   profile_history_rows_list = []
 
+  if not task_gcp_config.dataset_name:
```
RissyRan, line 703: No, we like to have metrics for each test at system level.

xlml/utils/metric.py
```diff
@@ -545,6 +545,48 @@ def get_gke_job_status(benchmark_id: str) -> bigquery.JobStatus:
   return bigquery.JobStatus.FAILED
 
 
+def get_gke_job_status(
```
RissyRan, line 548: Have you tested in both cases?

RissyRan (commented):
We discussed this offline as well. Separate it into 2 PRs: GKE support & timedelta change. Thanks Pei!

xlml/utils/metric.py
zpcore: Yes, tested for both Will's GKE task and torchbench with postprocess.

xlml/apis/gcp_config.py
zpcore: I didn't even notice that.  Updated to force the dataset_name

xlml/apis/task.py
```diff
@@ -249,9 +249,7 @@ def run_model(
 
     return tpu.ssh_tpu.override(
         task_id="run_model",
-        execution_timeout=datetime.timedelta(
-            minutes=self.task_test_config.time_out_in_min
-        ),
+        execution_timeout=self.task_test_config.time_out_in_min,
```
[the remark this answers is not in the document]
zpcore: Obsoleted. Updated in previous PR.

xlml/apis/gcp_config.py
zpcore: Do we still need the datasheet name even though we don't store any data? For example, the xlml test?

zpcore: Per offline discussion, we need the name.

will-cromar (approved):

PR #268, opened 2024-04-25
update gke cluster with resource from cloud-ml-benchmarking and add a…

zpcore:
…100/L4 resource node pool

# Description

1. Add resource from project `cloud-ml-benchmarking`;
2. Add node pool for A100 and L4;
3. Add node pool for V100 using `cloud-ml-benchmarking`

# Tests
Checked with `terraform plan`.
(No test)

# Checklist


deployment/clusters.tf
```diff
@@ -36,6 +52,27 @@ resource "google_container_node_pool" "primary" {
   }
 }
 
+resource "google_container_node_pool" "benchmark" {
```
mbzomowski, line 55: Can you break the node pools down into modules? E.g., like was done here https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/deployment/modules/composer_env/main.tf for google_composer_environment

You can create another directory under https://github.com/GoogleCloudPlatform/ml-auto-solutions/tree/master/deployment/modules, probably something like `node_pool`, and add a `main.tf` and `variables.tf` file, and create a module with those?

main.tf could probably be something like:

```
resource "google_container_node_pool" "node_pool" {
  name = var.node_pool_name
  project = var.node_pool_project
  location = var.node_pool_location
  ...etc
}
```

Then replace what you have here with

```
module "node_pool" {
  node_pool_name = "benchmark-pool"
  etc
}
```

Breaking down the terraform configs like this makes them more modular and easy to use.

zpcore: Check if we want to keep `google_container_cluster` still inside cluster.tf but move all `google_container_node_pool` under deployment/modules/nodepool/...?

mbzomowski: yeah that would be good tbh, since you're adding another cluster. You can do `google_container_cluster` and `google_container_node_pool` in separate modules, I think putting them together would get a bit complicated.

zpcore: Sounds good to me, thanks!

zpcore: Do you know why `terraform plan` can't detect my changes after moving pool related to the new folder `deployment/modules/nodepool`? It only reported node pool got destroyed because I removed node pool from clusters.tf file into the new location.

zpcore: I moved cluster under deployment/modules/cluster_node_pool/ also since the cluster and pool share the same variables like `cluster, location, project` etc..

mbzomowski: can you share the results of `terraform plan`?

deployment/modules/cluster_node_pool/main.tf
```diff
@@ -0,0 +1,288 @@
+resource "google_container_cluster" "gpu-uc1" {
```
mbzomowski, line 1: Sorry for the misunderstanding here - I think google_container_cluster should go under its own module, not under `cluster_node_pool`. Something like `deployment/modules/gke_cluster`.

deployment/clusters.tf
mbzomowski, line 1: Wait, why was clusters.tf deleted? clusters.tf should remain, but will just be able to use the modules you create under `deployment/modules` to create the same resources that were already being created there

zpcore: Sorry, I am a bit confused. Do you mean we leave everything in `clusters.tf` as it is but we have extra module in `deployment/modules/cluster_node_pool/`? In this case, which item should we add in `deployment/modules/cluster_node_pool/{main/variables}.tf`? I am trying to figure out the configuration boundry between those files.

PR #273, opened 2024-05-02
Add llama2-13b and gemma-7b

yeandy:
# Description

- Generalize the DAG creation function to support other models.
- Add llama2-13b, add gemma-7b models.
- Add separate DAGs for nightly and stable maxtext/jax/libtpu installs.

# Tests
Tested in my own dev Composer environment: http://shortn/_IXWgZp6KN3
Confirmed that metrics are also uploaded to BigQuery dev_benchmark_dataset.

# Checklist


yeandy:
R: @JoeZijunZhou

dags/inference/maxtext_inference.py
```diff
@@ -35,14 +36,26 @@
...
+      },
+      "llama2-13b": {
+          "per_device_batch_sizes": [1, 2],
+          "checkpoint": "gs://inference-benchmarks/models/llama2-13b/2024-04-25-14-01/param-only-decode-ckpt-maxtext/checkpoints/0/items",
+          "maxtext_logs": "gs://inference-benchmarks/models/llama2-13b/2024-04-25-14-01/",
+          "tokenizer": "tokenizer.llama2",
+      },
+      "gemma-7b": {
+          "per_device_batch_sizes": [1, 2, 4],
+          "checkpoint": "gs://inference-benchmarks/models/gemma-7b/2024-04-25-14-01/param-only-decode-ckpt-maxtext/checkpoints/0/items",
```
JoeZijunZhou, line 53: Will the model checkpoints be updated? Do we need to test the latest model checkpoints?

JoeZijunZhou (approved):
LGTM!

dags/inference/maxtext_inference.py
yeandy: These checkpoints are based on the PyTorch checkpoint found here `gs://maxtext-llama/llama2-7b/meta-ckpt`. (Which I pulled from existing MaxText end-to-end [tests](https://github.com/google/maxtext/blob/main/end_to_end/tpu/test_llama2_7b.sh#L25). My understanding is that this checkpoints were obtained from Meta. Though not sure if it was from https://github.com/meta-llama/llama, or HuggingFace. )

What I did was do a PyTorch->MaxText conversion, and then converted that rolled (scanned) version to an unnrolled (unscanned) version. That became the final checkpoint path that we see in the code now. This is to reduce time doing conversions every time. I also wrote an internal document on how to do this.

The checkpoints can be updated if the model weights change or we want to use a different fine-tuned version. Otherwise, we can keep it. @JoeZijunZhou  On this note, when you did your testing, did you download from https://github.com/meta-llama/llama? Also did you use the [chat or non-chat](https://github.com/meta-llama/llama/blob/main/download.sh#L36-L45) versions?

JoeZijunZhou: I tested with both `gs://maxtext-llama/llama2-7b/meta-ckpt` and non-chat ckpt from `https://github.com/meta-llama/llama`. Do you know if the source would change or not?

yeandy: Do you mean would the source checkpoints change? Not sure, but even if they do get updated versions, I don't think it should matter, at least on performance.

yeandy:
@vipannalla PTAL, thanks!

vipannalla:
Looks good, few minor comments

yeandy:
> Looks good, few minor comments

Hmm, I don't think they got published.

dags/inference/configs/maxtext_inference_gce_config.py
```diff
@@ -56,7 +58,7 @@ def get_maxtext_inference_nightly_config(
       "source .env/bin/activate",
       # Setup MaxText & JetStream
       f"cd maxtext && bash setup.sh MODE={test_mode.value} && cd ..",
-      "cd JetStream/benchmarks && pip install -r requirements.in",
```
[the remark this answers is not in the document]
yeandy: I had previously only installed the benchmark-specific dependencies https://github.com/google/JetStream/tree/main/benchmarks#install-dependencies. Hence the `cd JetStream/benchmarks && pip install -r requirements.in`

But I never installed the main dependencies https://github.com/google/JetStream/blob/main/README.md#setup. I think it worked before due to all the required dependencies installed when I installed maxtext. But some Jetstream-specific ones were introduced, which required explicit install of JetStream dependencies.

Is there a different path you're referring to?

dags/inference/configs/maxtext_inference_gce_config.py
```diff
@@ -65,44 +67,21 @@ def get_maxtext_inference_nightly_config(
...
+      f"export UNSCANNED_CKPT_PATH={model_configs['checkpoint']}",
+      f"export TOKENIZER_PATH=assets/{model_configs['tokenizer']}",
       "export LOAD_PARAMETERS_PATH=${UNSCANNED_CKPT_PATH}",
       "export MAX_PREFILL_PREDICT_LENGTH=1024",
       "export MAX_TARGET_LENGTH=2048",
-      "export MODEL_NAME=llama2-7b",
+      f"export MODEL_NAME={model_configs['model_name']}",
       "export ICI_FSDP_PARALLELISM=1",
       "export ICI_AUTOREGRESSIVE_PARALLELISM=-1",
       "export ICI_TENSOR_PARALLELISM=1",
```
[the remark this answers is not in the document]
yeandy: Updated the code to be able to configure the ici related params. 

Rerunning tests now.

dags/inference/maxtext_inference.py
yeandy: Do you mean if there's a way to always use the latest checkpoint from `gs://maxtext-llama/llama2-7b/meta-ckpt` (which is a copy from Meta) or directly from Meta (via HuggingFace, or https://github.com/meta-llama/llama)? 

If we point our path to `gs://maxtext-llama/llama2-7b/meta-ckpt` (and add back the conversion process), this relies on the assumption that the checkpoint there is maintained to be the most current. Will need to look into the ownership there, and guarantees (if any) on it being the latest.

I think if we source directly from Meta, we will be able to pull the latest checkpoint? But I haven't yet tested E2E feasibility, as it requires some manual steps (permissions, etc.) to even download the checkpoint. So not sure if/how it could be automated. 

In general, do we expect the llama2 checkpoints to be updated frequently from Meta? It may not be ideal, but if we don't expect frequent updates to the original weights, then would using this fixed path be good for now?

dags/inference/maxtext_inference.py
```diff
@@ -35,14 +36,32 @@
 ) as dag:
   test_name_prefix = "maxtext-inference"
   test_models = {
-      "llama2-7b": "llama2_7b",
-      # "mistral": "test_mistral",
-      # "gemma-2b": "gemma/2b/test_gemma",
-      # "gpt3": "test_gpt3",
+      "llama2-7b": {
+          "per_device_batch_sizes": [1, 2, 4],
```
vipannalla, line 40: FYI, the batch_sizes (and the ici_parallelism settings below) are tied to the hardware used. i.e, if we use v5p or v6e, these will be different. Don't need to update this CL, but something to remember for future CLs.

vipannalla (approved):
LGTM!

dags/inference/maxtext_inference.py
yeandy: Ack.

PR #279, opened 2024-05-07
Refactor `TpuQueuedResourceTask` into  a function

will-cromar:
# Motivation

This PR starts the process of replacing `BaseTask`.

`BaseTask` made more sense in earlier versions of the code. If you look at the ABC definition for `BaseTask` now, its implementations take some arbitrary number of arguments, and they're required to have a `run()` impl that takes no arguments: https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/ad0aad7c97ca74b42823714c58029e5352559fb2/xlml/apis/task.py#L29-L39

This is a function with extra steps. So, we should make each task type a function. This would reduce the number of assumptions we push down onto client teams and bring them closer to Airflow, making it easier to implement isolated features. At this point, we're not really doing anything with `Task`s that is helpful. And, the name `Task` conflicts with Airflow's own terminology for `@task`s, introducing even more confusion.

There's been persistent (and understandable) confusion about how to use `Task`. For example, extra entrypoints like `run_with_startup_script` break with the the `BaseTask` interface. This ends up being fine, because we never use BaseTask as an interface.

# Description

- Replace `TpuQueuedResourceTask(...).run()` with `run_queued_resource_test(...)`.
  - Inlined some code in functions that were never re-used.
- Remove unused variants of `run_*`.
- Remove some obsolete DAGs

# Tests

Screenshots:

<details>

Local run of PT/XLA test gets same results:

image

Graph structure is unchanged so history is preserved (dev env):

image

Test runs in dev env across a couple more dags:

image

image

</details>

The list of code owners is insane. cc @RissyRan @mbzomowski @jonb377 to take an early look.


dags/pytorch_xla/demo.py
```diff
@@ -0,0 +1,24 @@
+import datetime
```
RissyRan, line 1: Nit: header for the file,

xlml/apis/task.py
```diff
@@ -300,6 +175,19 @@ def clean_up(self, queued_resource: airflow.XComArg) -> DAGNode:
         queued_resource
     )
 
+  with TaskGroup(
```
RissyRan, line 178: Do you have a test I could take a quick look about graphs?

RissyRan (commented):
Thanks Will! Two questions at high level:

1) Why do we remove the sweep examples? If so, let's get it reviewed by Raymond (the author added the example).
2) I assume we will have follow up refactors for all other tasks? i.e. gke, xpk on both TPU an GPU?

will-cromar:
> 1. Why do we remove the sweep examples? If so, let's get it reviewed by Raymond (the author added the example).

These DAGs are currently unused. The real tests under `multipod` have been moved to XPK. I checked with Raymond and others before deleting them.

> 2. I assume we will have follow up refactors for all other tasks? i.e. gke, xpk on both TPU an GPU?

Yeah. Other tasks can be done in follow ups, either as ramp-up tasks (most will be easier than the QR task) or as I have time to do clean-up.

xlml/apis/task.py
will-cromar: I still need to do more testing. Since this is a big change, I'll will try deploying it to the dev environment and posting links/screenshots.

dags/pytorch_xla/demo.py
will-cromar: Oops, this file is from a live demo I did a few weeks ago. Will remove.

xlml/apis/task.py
will-cromar: Ran some tests in the dev env and posted screenshots.

will-cromar:
In an earlier draft, I tried to use `@task_group`, but it has subtly different behaviors I couldn't get exactly right, so I largely reverted the logic to what it was before. It now just lives in a simplified function that is more readable end-to-end.

RissyRan (approved):
Thanks! LGTM!

will-cromar:
Since this is a no-op, I'm going to submit past code owners.

If I am wrong and this is not a no-op, please let me know and we can roll this PR back.

PR #293, opened 2024-05-23
Delete unused JAX pod smoke tests

will-cromar:
Confirmed with JAX team that they no longer use these. The actual test case was initializing JAX over a small slice. We would know from other sources (ie almost every other test in this repo) if JAX did not work on slices anymore.


sshahrokhi (approved):
Thanks Will for confirming.

allenwang28 (approved):

PR #294, opened 2024-05-23
Enable torchbench nightly tests on torch_xla2

zpcore:
# Description

Torchbench daily run on torch-xla2.

I comment out tests on majority of GPU platforms to avoid competing resources with other runs.

Note: This PR depends on https://github.com/pytorch/xla/pull/7126/. Wait for that PR to merge before this one.

# Tests

The dags should only impact torch_xla2

https://60bb71acc6784780b3bbdbae4ffa636f-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla2-torchbench/grid

# Checklist

### DIFF
--- a/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
+++ b/dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
@@ -35,12 +35,11 @@ class VERSION_MAPPING:
 
   class NIGHTLY(enum.Enum):
     TORCH_XLA_TPU_WHEEL = "https://storage.googleapis.com/pytorch-xla-releases/wheels/tpuvm/torch_xla-nightly-cp310-cp310-linux_x86_64.whl"
-    TORCH_XLA_CUDA_WHEEL = "https://storage.googleapis.com/pytorch-xla-releases/wheels/cuda/12.1/torch_xla-nightly-cp38-cp38-linux_x86_64.whl"
+    TORCH_XLA_CUDA_WHEEL = "https://storage.googleapis.com/pytorch-xla-releases/wheels/cuda/12.1/torch_xla-nightly-cp310-cp310-linux_x86_64.whl"
     TORCH = "torch"
     TORCHVISION = "torchvision"
     TORCHAUDIO = "torchaudio"
-    # TODO(@piz): update to xla:nightly_3.10_cuda_12.1 once available
-    TORCH_XLA_GPU_DOCKER = "us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.8_cuda_12.1"
+    TORCH_XLA_GPU_DOCKER = "us-central1-docker.pkg.dev/tpu-pytorch-releases/docker/xla:nightly_3.10_cuda_12.1"
     TORCH_INDEX_CPU_URL = "https://download.pytorch.org/whl/nightly/cpu"
     TORCH_INDEX_CUDA_URL = "https://download.pytorch.org/whl/nightly/cu121"
     TORCH_REPO_BRANCH = "-b main"
@@ -95,7 +94,10 @@ def get_version_mapping(test_version):
 
 
 def set_up_torchbench_tpu(
-    model_name: str = "", test_version: VERSION = VERSION.NIGHTLY
+    model_name: str = "",
+    test_version: VERSION = VERSION.NIGHTLY,
+    use_startup_script: bool = False,
+    use_xla2: bool = False,
 ) -> Tuple[str]:
   """Common set up for TorchBench."""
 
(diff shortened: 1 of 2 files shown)


dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
```diff
@@ -242,6 +273,17 @@ def get_nvidia_driver_install_cmd(driver_version: str) -> str:
     )
     return nvidia_driver_install
 
+  install_torch_xla2_dependency = (
+      (
+          # TODO(piz): torch_xla2 only support nightly test at this time.
+          "pip install torch_xla[pallas] -f https://storage.googleapis.com/jax-releases/jax_nightly_releases.html -f https://storage.googleapis.com/jax-releases/jaxlib_nightly_releases.html",
```
will-cromar, line 279: What's the reason for installing `torch_xla[pallas]`? `torch_xla2` doesn't depend on the main package.

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
```diff
@@ -242,6 +273,17 @@ def get_nvidia_driver_install_cmd(driver_version: str) -> str:
...
+          "cd /tmp/xla/experimental/torch_xla2/",
```
will-cromar, line 280: Do you need to clone this? You should be able actually install directly from git: https://pip.pypa.io/en/stable/topics/vcs-support/#url-fragments

That's one advantage of a pure python project. That wouldn't work for the main `torch_xla` package.

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
```diff
@@ -115,6 +117,17 @@ def model_install_cmds(output_file=None) -> str:
       return f"python install.py --continue_on_fail {pipe_file_cmd}"
     return f"python install.py models {model_name} {pipe_file_cmd}"
 
+  install_torch_xla2_dependency = (
+      (
+          # TODO(piz): torch_xla2 only support nightly test at this time.
+          "pip install torch_xla[pallas] -f https://storage.googleapis.com/jax-releases/jax_nightly_releases.html -f https://storage.googleapis.com/jax-releases/jaxlib_nightly_releases.html",
+          "cd ~/xla/experimental/torch_xla2/",
+          "pip install -e .",
```
will-cromar, line 125: Should this be `.[tpu]` / `torch_xla2[tpu]`?

zpcore: I should mention this PR: https://github.com/pytorch/xla/pull/7126/. I am trying not to reinstall libtpu to an version that doesn't match with the torch_xla[pallas].
`pip install -e .` will not install jax/jaxlib. Instead we use the jax installation from torch_xla[pallas].

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
zpcore: Oh, good to know we can install directly.

We still need scripts under xla/benchmarks/..., so we need to clone the repo anyway.

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
zpcore: Just want to use the jax, libtpu version from what is specified in torch_xla uniformly. This helps us use the same script to test on both torch_xla and torch_xla2. WDYT?

will-cromar: Let's let them specify JAX versions separately. Even if it works now, it will break in the future. To use the same VM with two different sets of packages, I recommend creating a venv: https://docs.python.org/3/library/venv.html

zpcore:
Isolating the libtpu dependencies between torch_xla and torch_xla2 for benchmarking is added in https://github.com/pytorch/xla/pull/7144. After the change, we can install `pip install -e .[cpu/tpu/cuda]` for torch_xla2 without concerning libtpu for torch_xla when do benchmarking.

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
zpcore: Update to use .[tpu] after isolated the torch_xla and torch_xla2.

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
zpcore: Thanks, I will add creating the env for different test as a TODO. Currenly I have isolated the torch_xla and torch_xla2 benchmarking dependencies in https://github.com/pytorch/xla/pull/7144. We don't install torch_xla[pallas] in the script anymore.

However, one thing I want to mention is that using the same libtpu version between torch_xla and torch_xla2 can give us a fail comparsion between those two.

zpcore:
Need to do a refactor to git repo pytorch/xla/benchmarks/*.py scripts to completely get rid of (or add import condition to) the global import of `torch_xla` related items. 

Marks this as a TODO since this will depend on changes on other PRs.

dags/pytorch_xla/configs/pytorchxla_torchbench_config.py
zpcore: Based on offline discussion, we uninstaill the libtpu from torch_xla and reinstall based on torch_xla2. We don't care about the libtpu version difference in this case.

will-cromar (approved):

PR #298, opened 2024-05-29
Retry GKE `stream_logs` when status is unkown

will-cromar:
Currently this either returns success if there is no exception and no status, or it starts a new attempt if there is an exception.

# Tests

Tested locally. Jobs are failing if the sensor retries after the `job_create_timeout`, even if they're within the test config's `timeout`. Needs follow-up.

### DIFF
--- a/xlml/utils/gke.py
+++ b/xlml/utils/gke.py
@@ -147,7 +147,15 @@ def run_job(
 
       # Wait for pods to complete, and exit with the first non-zero exit code.
       for f in concurrent.futures.as_completed(futures):
-        exit_code = f.result()
+        try:
+          exit_code = f.result()
+        except kubernetes.client.ApiException as e:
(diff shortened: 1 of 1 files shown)


zpcore:
Thanks, can we also specify the timeout_seconds=0 in stream() (https://stackoverflow.com/questions/61611261/kubernetes-python-client-watch-stops-returning-results-doesnt-fail) to prevent task exit when logging take longer than 5 mins to make a update?

will-cromar:
> Thanks, can we also specify the timeout_seconds=0 in stream() (https://stackoverflow.com/questions/61611261/kubernetes-python-client-watch-stops-returning-results-doesnt-fail) to prevent task exit when logging take longer than 5 mins to make a update?

Let's adjust the timeouts in a follow-up.

allenwang28 (approved):

PR #305, opened 2024-06-04
Add tests for MaxText automatic profile upload to Vertex AI Tensorboard

SurbhiJainUSC:
# Description

This PR adds a test to check that MaxText profile has been uploaded to Vertex AI Tensorboard instance.

# Tests

- Ran local airflow server to verify that the DAG is up and running
- The tests are failing currently for a known reason and we are working on fixing it

# Checklist

### DIFF
--- a/dags/multipod/maxtext_profiling_vertex_ai_tensorboard.py
+++ b/dags/multipod/maxtext_profiling_vertex_ai_tensorboard.py
@@ -1,0 +1,62 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+A DAG to test MaxText Automatic Profile Upload to Vertex AI Tensorboard.
+"""
+import datetime
+from airflow import models
+from dags import composer_env, test_owner, gcs_bucket
+from dags.vm_resource import TpuVersion, Zone, DockerImage
+from dags.multipod.configs import gke_config
+from dags.multipod.configs.common import SetupMode
+
+# Run once a day at 6 am UTC (10 pm PST)
+SCHEDULED_TIME = "0 6 * * *" if composer_env.is_prod_env() else None
+
+with models.DAG(
+    dag_id="maxtext_profiling_vertex_ai_tensorboard",
+    schedule=SCHEDULED_TIME,
+    tags=["multipod_team", "maxtext", "stable", "nightly", "vertex_ai"],
+    start_date=datetime.datetime(2024, 6, 1),
+    catchup=False,
+    concurrency=2,
+) as dag:
+    base_output_directory = (
+        f"{gcs_bucket.BASE_OUTPUT_DIR}/maxtext_vertex_ai_tensorboard"
+    )
+    dataset_path = gcs_bucket.MAXTEXT_DIR
+    docker_images = [
+        (SetupMode.STABLE, DockerImage.MAXTEXT_TPU_JAX_STABLE),
+        (SetupMode.NIGHTLY, DockerImage.MAXTEXT_TPU_JAX_NIGHTLY),
+    ]
+
+    for mode, image in docker_images:
+        profiling_in_vertex_ai_tb_cmds = (
+            f"export RUN_NAME=vertex_ai_{mode.value}_$(date +%Y-%m-%d-%H-%M-%S)",
+            "python3 MaxText/train.py MaxText/configs/base.yml"
+            f" run_name=$RUN_NAME base_output_directory={base_output_directory}"
+            f" dataset_path={dataset_path} profiler=xplane steps=10",
+            "gsutil ls gs://cloud-ai-platform-*/tensorboard-*/$EXPERIMENT_NAME",
+        )
+        profiling_in_vertex_ai_tb_test = gke_config.get_gke_config(
+            tpu_version=TpuVersion.V4,
+            tpu_cores=8,
+            tpu_zone=Zone.US_CENTRAL2_B.value,
+            time_out_in_min=240,
+            test_name=f"profiling-vertex-ai-tensorboard-{mode.value}",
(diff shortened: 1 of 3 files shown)


xlml/utils/xpk.py
```diff
@@ -63,18 +64,24 @@ def run_workload(
...
-        (
-            f"python {tmpdir}/xpk/xpk.py workload create"
-            f" --cluster={cluster_name} --workload={workload_id}"
-            f" --command='{run_cmds}' --device-type={accelerator_type}"
-            f" --{multi_keyword}={num_slices} --docker-image={docker_image}"
-            f" --project={cluster_project} --zone={zone}"
-            f" --env {metric_config.SshEnvVars.GCS_OUTPUT.name}={gcs_path}"
-            " --restart-on-user-code-failure"
-        ),
+        "pip install cloud-accelerator-diagnostics",
```
bvandermoon, line 83: Should this be behind the ```use_vertex_tensorboard``` flag also? Or is it needed in other places?

dags/multipod/maxtext_profiling_vertex_ai_tensorboard.py
```diff
@@ -0,0 +1,62 @@
...
+      (SetupMode.NIGHTLY, DockerImage.MAXTEXT_TPU_JAX_NIGHTLY),
+  ]
+
+  for mode, image in docker_images:
+    profiling_in_vertex_ai_tb_cmds = (
+        f"export RUN_NAME=vertex_ai_{mode.value}_$(date +%Y-%m-%d-%H-%M-%S)",
+        "python3 MaxText/train.py MaxText/configs/base.yml"
+        f" run_name=$RUN_NAME base_output_directory={base_output_directory}"
+        f" dataset_path={dataset_path} profiler=xplane steps=10",
+        "gsutil ls gs://cloud-ai-platform-*/tensorboard-*/$EXPERIMENT_NAME",
```
bvandermoon, line 51: Is ```gsutil ls``` here enough to pass/fail the test based on the presence of the file? Does it need an exception thrown if it's not present?

dags/multipod/maxtext_profiling_vertex_ai_tensorboard.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    profiling_in_vertex_ai_tb_test = gke_config.get_gke_config(
+        tpu_version=TpuVersion.V4,
```
bvandermoon, line 54: Any reason for only running on v4?

dags/multipod/maxtext_profiling_vertex_ai_tensorboard.py
SurbhiJainUSC: This command will throw the exception: `CommandException: One or more URLs matched no objects.` when `$EXPERIMENT_NAME` object is not created by Vertex AI Tensorboard uploader inside `cloud-ai-platform-*` bucket and will fail the test. When the object is created, no exception is thrown and the test passes.

xlml/utils/xpk.py
SurbhiJainUSC: It is not needed in other places. I can move it behind the flag for readability.

dags/multipod/maxtext_profiling_vertex_ai_tensorboard.py
SurbhiJainUSC: I think this test is not specific to any TPU version, so thought of running it only on v4-8. Do you think we should add it to v4-16 and v5s? Imo it doesn't add any value, wdyt @bvandermoon?

dags/multipod/maxtext_profiling_vertex_ai_tensorboard.py
jonb377, line 51: Where does `EXPERIMENT_NAME` come from? xpk?

bvandermoon: Great, thanks Surbhi

dags/multipod/maxtext_profiling_vertex_ai_tensorboard.py
bvandermoon: I agree that in general the Vertex AI TB uploader shouldn't be impacted by TPU versions. But it would still be good to test on different versions like v4-16 and v5. Just to make sure that single slice/Multislice are both working properly, nothing in the environment is causing issues, etc.

xlml/utils/xpk.py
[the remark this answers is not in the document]
jonb377: Chatted offline - this is a dependency of `xpk` to create Vertex tensorboards. Since the package will update frequently, we'll keep a direct `pip install -U` here. Thanks @SurbhiJainUSC!

dags/multipod/maxtext_profiling_vertex_ai_tensorboard.py
SurbhiJainUSC: I have added the tests for single slice v4-8 and single/multislice v4-16. Currently, we have support for Vertex AI in the shared clusters for v4. In future, we can extend these tests for v5 when we add the Vertex AI support to that cluster.

xlml/utils/xpk.py
```diff
@@ -63,19 +64,30 @@ def run_workload(
...
+          f"git clone https://github.com/google/xpk {tmpdir}/xpk",
+          "pip install -U google-cloud-aiplatform cloud-accelerator-diagnostics",
+          workload_create_cmd,
+      )
+    else:
+      cmds = (
+          "set -xue",
+          f"git clone https://github.com/google/xpk {tmpdir}/xpk",
+          workload_create_cmd,
+      )
```
RissyRan, line 90: Could we rewrite something like this? So we have slightly concise code block.

```
vertext_dependency = "pip install -U google-cloud-aiplatform cloud-accelerator-diagnostics" if use_vertex_tensorboard else None

      cmds = (
          "set -xue",
          f"git clone https://github.com/google/xpk {tmpdir}/xpk",
          vertext_dependency,
          workload_create_cmd,
      )
```

SurbhiJainUSC: This will not work for tests where vertex_dependency="". The error will be thrown as https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/utils/xpk.py#L81 would result in `git clone https://github/com/google/xpk tmp54/xpk;;python...`. I will update `cmds` to a list instead of tuple to improve the readability of the code.

RissyRan: Ah, I see. Makes sense. Thanks!

RissyRan (approved):
LGTM! Thanks!

bvandermoon (approved):

jonb377 (approved):
LGTM - thanks Surbhi!

PR #337, opened 2024-06-20
Multi-Model MLPerf4.0 Reproduce on A3

jyj0w0:
# Description

Add Multi-model MLPerf4.0 reproduce automation.
Also bring network config for GPU to front. 

# Tests
https://screenshot.googleplex.com/3LYoaodHP77ssMz
# Checklist


dags/inference/configs/trt_llm_mlperf_v40_config.py
```diff
@@ -0,0 +1,153 @@
...
+      "sed -i '29i\ARCH=x86_64' Makefile.docker",
+      "sudo usermod -a -G docker $USER",
+      # Build and launch a docker container
+      "make prebuild DOCKER_DETACH=1",
+      "make docker_add_user",
+      f"make launch_docker DOCKER_NAME={docker_container_name} DOCKER_ARGS='-d'",
+  )
+
+  jsonl_output_path = "metric_report.jsonl"
+  jsonl_converter_py_lines = (
```
yeandy, line 78: Nice! I was thinking about doing something like this in the past, but didn't get around to it. I will likely borrow this in the future 😄

dags/inference/configs/trt_llm_mlperf_v40_config.py
```diff
@@ -0,0 +1,153 @@
...
+      "mv models/Llama2/fp8-quantized-ammo/llama2-70b-chat-hf-tp2pp1-fp8/ models/Llama2/fp8-quantized-ammo/llama2-70b-tp2pp1-fp8/",
+      "git clone https://github.com/mlcommons/inference_results_v4.0",
+      "cd /scratch/inference_results_v4.0/closed/Google",
+      "export MLPERF_SCRATCH_PATH=/scratch",
+      "cp /scratch/inference_results_v4.0/closed/{NVIDIA,Google}/Makefile.docker",
+      "sed -i '27i\ARCH=x86_64' Makefile",
+      "sed -i '29i\ARCH=x86_64' Makefile.docker",
+      "sudo usermod -a -G docker $USER",
+      # Build and launch a docker container
+      "make prebuild DOCKER_DETACH=1",
```
yeandy, line 72: How long does build take? I remember it took a while in the past.

dags/inference/configs/trt_llm_mlperf_v40_config.py
```diff
@@ -0,0 +1,153 @@
...
+      "sudo mkfs.ext4 -F /dev/md0",
+      "sudo mkdir -p /scratch",
+      "sudo mount /dev/md0 /scratch",
+      "sudo chmod a+w /scratch",
+      "cd /scratch",
+      # Prepare data
+      "mkdir -p models",
+      "mkdir -p preprocessed_data",
+      "gsutil -m cp -n -r gs://tohaowu/mlpinf-v40/models/Llama2 models/",
+      "gsutil -m cp -n -r gs://tohaowu/mlpinf-v40/preprocessed_data/open_orca preprocessed_data/",
```
yeandy, line 62: Can you please explain how this data was generated? Was it generated for last round?

I think the data (and models) should be the same for this upcoming round. So should we rename this, or make a copy of the data into a new GCS bucket to be more clear?

dags/inference/configs/trt_llm_mlperf_v40_config.py
```diff
@@ -0,0 +1,153 @@
...
+      "      writter.write(data)",
+      "if __name__ == '__main__':",
+      "  convert_to_jsonl(metadata_log_path, sys.argv[1])",
+  )
+  py_script = "\n".join(jsonl_converter_py_lines)
+  make_jsonl_converter_cmd = f'echo "{py_script}" > jsonl_converter.py'
+
+  docker_cmds = (
+      "make link_dirs",
+      "make build BUILD_TRTLLM=1",
```
yeandy, line 104: how long does this take?

dags/inference/configs/trt_llm_mlperf_v40_config.py
```diff
@@ -0,0 +1,153 @@
...
+      f'make run RUN_ARGS="--benchmarks={model_configs["model_name"]} --scenarios=Offline"',
```
yeandy, line 105: This code has integrations with loadgen, right? So it produces actual MLPerf outputs such as `mlperf_log_summary.txt`, etc.  I'm wondering, should we parse that file instead? 

Can you explain what `/scratch/inference_results_v4.0/closed/Google/build/logs/*/*/*/*/metadata.json` contains? Is it similar to `mlperf_log_summary.txt`? If `metadata.json` is similar, and simply contains a cleaned version of `mlperf_log_summary.txt`, then we can use it. However, if it contains a subset of info, then we can consider (in a new PR) more logic to parse `mlperf_log_summary.txt`.

dags/inference/configs/trt_llm_mlperf_v40_config.py
jyj0w0: Yes it took a while - around 1h

dags/inference/configs/trt_llm_mlperf_v40_config.py
jyj0w0: @tohaowu I think this data is MLPerf4.0's round, is this same as MLPerf4.1's ? Thanks!

dags/inference/configs/trt_llm_mlperf_v40_config.py
jyj0w0: ~ 1h build trtllm will take long time.

dags/inference/configs/trt_llm_mlperf_v40_config.py
jyj0w0: Yes it contains subset of `mlperf_log_summary.txt`, like without the "additional stats". I think we can use it for now since it is much easier to parse to `jsonl` format and switch to `mlperf_log_summary.txt` once the automation is mostly set up.

dags/inference/configs/trt_llm_mlperf_v40_config.py
yeandy: Is there a way to use a saved a snapshot, instead of building fresh every time? Just want to make the entire e2e benchmarks run faster 😄

dags/inference/configs/trt_llm_mlperf_v40_config.py
yeandy: Great, thanks for the clarification! Yeah, we can use `metadata.json` for now, as long as we extract as much relevant info as possible from that file into BigQuery.

dags/inference/configs/trt_llm_mlperf_v40_config.py
yeandy: Similar to the other comment, is there a way to use saved images/snapshots/etc. to shorten this time? If not, then so be it, but just want to strike a balance between using the most up-to-date software, vs, executing reasonably quick runs.

dags/inference/configs/trt_llm_mlperf_v40_config.py
jyj0w0: Good question @tohaowu do we have to build each time ? Thanks!

tohaowu: we only need to build once

jyj0w0: Yeah but I don't think there is any snapshot we can download so we still need to build every time a instance starts instead of downloading a snapshot and use for all instances. @tohaowu please correct me if I am wrong thanks! @yeandy

dags/vm_resource.py
```diff
@@ -29,6 +29,12 @@
 V5E_BM_SUBNETWORKS = f"{BM_NETWORKS}/regions/us-west1/subnetworks/mas-test"
 V5P_BM_SUBNETWORKS = f"{BM_NETWORKS}/regions/us-east5/subnetworks/mas-test"
 
+INFERENCE_NETWORK_PREFIX = "projects/cloud-tpu-inference-test"
+INFERENCE_NETWORKS = (
+    f"{INFERENCE_NETWORK_PREFIX}/global/networks/yijiaj-mas-test"
```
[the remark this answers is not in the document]
jyj0w0: I created one and it passed to create the instance https://screenshot.googleplex.com/6ahs78Bj9AFkisU

dags/inference/configs/trt_llm_mlperf_v40_config.py
```diff
@@ -0,0 +1,153 @@
...
+      "  data['metrics'] = dict()",
+      "  with open(json_path, 'r') as file:",
+      "      metadatadata = json.load(file)",
+      "      for key in metadatadata:",
+      "          try:",
+      "              float(metadatadata[key])",
+      "              data['metrics'][key] = float(metadatadata[key])",
+      "          except:",
+      "              data['dimensions'][key] = metadatadata[key]",
+      "  with jsonlines.open(jsonl_path, 'w') as writter:",
```
yeandy, line 94: nit: writer

dags/inference/configs/trt_llm_mlperf_v40_config.py
yeandy, line 105: What batch size does this run? Do you plan to run with different permutations of batch size, number of chips, quantization, etc.? (If so, we can add in a always add in different PR). Just curious.

yeandy: Maybe it's already encoded in the `fp8-quantized-ammo/llama2-70b-chat-hf-tp2pp1-fp8` name. What does ammo and tp2pp1 mean?

yeandy (commented):
Can you please share (here or in a private message) the schema for the metrics and metadata you collect into BigQuery? 

We may need to align the naming with the schema that I have for the TPU side. But we can do this as a separate query in BigQuery.

jyj0w0:
> Can you please share (here or in a private message) the schema for the metrics and metadata you collect into BigQuery?
> 
> We may need to align the naming with the schema that I have for the TPU side. But we can do this as a separate query in BigQuery.

The result is https://screenshot.googleplex.com/QxBDTQ3kmcGcVU5, we can align the schema while querying since the metrics are from metadata.json, the format is firmed with MLPerf imo. One result for example ```    "effective_samples_per_query": 50160,
    "result_samples_per_second": 71.7945,
    "result_tokens_per_second": 20800.4```

yeandy:
Thanks! Yes, we can align schema later if we need to unify anything for the dashboard

dags/inference/configs/trt_llm_mlperf_v40_config.py
jyj0w0: For A3(and A3+) the config is settled to MLPerf4.0 tuned (batch size is [16 by default]( https://github.com/mlcommons/inference_results_v4.0/blob/ceef1eab947e15915d3208c9ce8c52ed418dd1f7/closed/Google/code/llama2-70b/tensorrt/builder.py#L58)) . For future use cases for G2 and A2 the parameters would be tuned so probably need some similar sweeping logic. 
The names are set here https://github.com/mlcommons/inference_results_v4.0/blob/ceef1eab947e15915d3208c9ce8c52ed418dd1f7/closed/Google/code/llama2-70b/tensorrt/builder.py#L53 @tohaowu might know more about what the names means.

yeandy (approved):

xlml/apis/test_config.py
```diff
@@ -101,13 +101,17 @@ class Gpu(Accelerator):
...
+    network: The network that a GPU will be a part of.
+    subnetwork: The subnetwork that a GPU will be a part of.
   """
 
   machine_type: str
   image_family: str
   count: int
   accelerator_type: str
   runtime_version: Optional[str] = None
+  network: str = 'global/networks/default'
```
RissyRan, line 113: Are those network and subnetwork used here?

jyj0w0: Looks `None` and `global/networks/default` are same https://screenshot.googleplex.com/8UxZNGXnwrXLDab - it was `global/networks/default` before so just put it here.

will-cromar: `None` makes more sense to me. The network and subnetwork are essentially meaningless if the GPU is in GKE since the network is determined by the cluster itself (not necessarily `default`).

will-cromar (approved):

PR #339, opened 2024-06-24
Add MaxText Llama2 to v5e perf tests

raymondzouu:
# Description

Add MaxText Llama2 7B, 13B, 70B to v5e perf tests

# Tests

Ran in local Composer env: https://screenshot.googleplex.com/BaPbwkw755Di5EJ

# Checklist

### DIFF
--- a/dags/multipod/maxtext_v5e_configs_perf.py
+++ b/dags/multipod/maxtext_v5e_configs_perf.py
@@ -32,11 +32,20 @@ with models.DAG(
     tags=["multipod_team", "maxtext", "stable", "nightly"],
     start_date=datetime.datetime(2024, 2, 19),
     catchup=False,
-    concurrency=10,
+    concurrency=25,
 ) as dag:
   # MaxText set up
   quantization_sweep = {"M_QUANTIZATION": ["", "int8"]}
-  model_configs = ["16b", "32b", "64b", "128b", "gpt3_175b"]
+  model_configs = [
+      "16b",
+      "32b",
+      "64b",
+      "128b",
+      "gpt3_175b",
+      "llama2_7b",
+      "llama2_13b",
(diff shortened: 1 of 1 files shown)


dags/multipod/maxtext_v5e_configs_perf.py
```diff
@@ -56,7 +65,7 @@
           dataset_name=metric_config.DatasetOption.XLML_DATASET,
           cluster_name=ClusterName.V5E_256_US_WEST_4_MULTISLICE_CLUSTER.value,
```
jonb377, line 66: How long do these tests run for? We currently allow jobs to queue in the cluster for 5 hours before timeout: https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/06efab3a05ea2d53135560b9a4a299ed0a1f34df/xlml/apis/task.py#L149-L151

With 32 workloads and no dependencies in the DAG (unless I'm missing some in the sweep), the timeout will occur at about the same time for all tasks, so we'd need all to complete within 5 hours.

jonb377 (approved):
LGTM! Let me know if the timeout will be an issue.

dags/multipod/maxtext_v5e_configs_perf.py
raymondzouu: Ah that's a good point. There's 64 workloads and each workload runs for 30 steps with the step times ranging from 2-15 sec. We have 2 pods and half of the workloads require the full 2 pods so then there are 48 workloads that will run sequentially. So it should take 30steps*10sec*48workloads = 14400sec = 240min for all of the workloads to run in the cluster. So hopefully they should finish within time based on that calculation :D but if not I can add some dependencies or lower the number of steps later.

PR #344, opened 2024-07-02
Fix check of bool str for quantize_kvcache

yeandy:
# Description

In our config, when were setting `quantize_kvcache` to `False`, we used the string "false". But the logic
```
'int8' if model_configs['quantize_kvcache'] else 'bf16'
```
produced "int8" because the string "false" is truthy.

When merged, need to update entries in BQ `quant_mode` is w-b16-kv-b16 but we accidentally set `kvcache_quant_dtype` to "int8"

# Tests
Test in Python shell.

# Checklist


dags/inference/configs/jetstream_benchmark_serving_gce_config.py
```diff
@@ -84,7 +84,7 @@ def get_config(
       "quantization": f"{model_configs['quantization']}",
       "quantize_kvcache": f"{model_configs['quantize_kvcache']}",
       "weight_quant_dtype": f"{model_configs['quantization'] if model_configs['quantization'] else 'bf16'}",
-      "kvcache_quant_dtype": f"{'int8' if model_configs['quantize_kvcache'] else 'bf16'}",
+      "kvcache_quant_dtype": f"{'int8' if model_configs['quantize_kvcache'].lower() == 'true' else 'bf16'}",
```
morgandu, line 87: I think right now we also support int4 for kvcache_quant_dtype

yeandy: Got it, thanks for the pointer. I updated this variable to `kv_quant_dtype`. PTAL

morgandu (approved):

PR #345, opened 2024-07-02
Update Mixtral BS

yeandy:
# Description

Update Mixtral batch size since 256 taking very long time. 

# Tests

# Checklist

### DIFF
--- a/dags/inference/maxtext_inference.py
+++ b/dags/inference/maxtext_inference.py
@@ -457,7 +457,7 @@ with models.DAG(
           "quant_mode": W_INT8_KV_INT8,
           "quantization": "int8",
           "quantize_kvcache": "true",
-          "per_device_batch_size": 258,
+          "per_device_batch_size": 128,
           "kv_quant_axis": "heads_and_dkv",
           "run_eval": True,
       },


vipannalla:
Higher batch_size is supposed to be faster given the same number of samples. How much slower is the current run? and how much faster is the run with batch_size = 128?

yeandy:
It takes about ~12 min for 128 batch on bf16. Haven't tested for int8.

Not sure why it's hitting timeout at 3 hours for int8 version. Let me try running int8 at 128.

dags/inference/maxtext_inference.py
```diff
@@ -457,7 +457,7 @@
           "quant_mode": W_INT8_KV_INT8,
           "quantization": "int8",
           "quantize_kvcache": "true",
-          "per_device_batch_size": 258,
```
RissyRan, line 460: Wait! we have already quantized for inference on MaxText?

yeandy: Yes!

RissyRan: Which implementation is this one? I don't think I enabled quantization for both matmul or megablox yet (at least for the MoE block). Or other implementation we are talking about here?

vipannalla:
I think this configuration is using the old for `for loop` implementation, which Ranran cleaned up recently. I don't think the current code support quantization right now.

PR #346, opened 2024-07-02
Add dependencies between v5e perf tests

raymondzouu:
# Description

Add dependencies between v5e perf tests

# Tests
 N/A

# Checklist

### DIFF
--- a/dags/multipod/maxtext_v5e_configs_perf.py
+++ b/dags/multipod/maxtext_v5e_configs_perf.py
@@ -32,7 +32,6 @@ with models.DAG(
     tags=["multipod_team", "maxtext", "stable", "nightly"],
     start_date=datetime.datetime(2024, 2, 19),
     catchup=False,
-    concurrency=25,
 ) as dag:
   # MaxText set up
   quantization_sweep = {"M_QUANTIZATION": ["", "int8"]}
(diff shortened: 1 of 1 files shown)


jonb377 (approved):
LGTM, thanks for the patch Raymond!

jonb377:
cc @RissyRan, we're hitting queueing timeouts in XPK tasks due to the volume of tests we're running.

I'm only aware of two options to address this:
1. Add dependencies between the tests, which this PR does. The downside is if one test fails/times out, downstream tests will be skipped.
2. Increase the queueing timeout. The downside here is that with a higher timeout, DAGs with multiple dependent XPK tests can take longer than 24 hours to execute.

Do you know of any other resolutions?

RissyRan:
> * Add dependencies between the tests, which this PR does. The downside is if one test fails/times out, downstream tests will be skipped.

In the 1st approach, will the last step succeed always? i.e. post-process or clean up. If so, the next test will not be skiped.

raymondzouu:
> > * Add dependencies between the tests, which this PR does. The downside is if one test fails/times out, downstream tests will be skipped.
> 
> In the 1st approach, will the last step succeed always? i.e. post-process or clean up. If so, the next test will not be skiped.

Oh yes the post-process should succeed regardless of model test failure, thanks!

PR #348, opened 2024-07-08
Increase number of dependencies in v5e maxtext perf tests

raymondzouu:
# Description

Increase number of dependencies in v5e maxtext perf tests due to workload start timeouts.

DAG dependencies now look like this https://screenshot.googleplex.com/7TfW2uPwC3xo9Rn

# Tests

N/A

# Checklist

### DIFF
--- a/dags/multipod/maxtext_v5e_configs_perf.py
+++ b/dags/multipod/maxtext_v5e_configs_perf.py
@@ -75,11 +75,10 @@ with models.DAG(
           sweep_params=quantization_sweep,
       )
 
-      num_tests = len(maxtext_sweep_gke_test)
-      for i in range(num_tests // 2):
-        (
-            maxtext_sweep_gke_test[i].run_with_run_name_generation()
-            >> maxtext_sweep_gke_test[
-                num_tests - 1 - i
-            ].run_with_run_name_generation()
-        )
+      chain_num = 4
+      prev = maxtext_sweep_gke_test[0].run_with_run_name_generation()
+      for i in range(1, len(maxtext_sweep_gke_test)):
+        curr = maxtext_sweep_gke_test[i].run_with_run_name_generation()
+        if i % chain_num != 0:
+          prev >> curr
+        prev = curr


dags/multipod/maxtext_v5e_configs_perf.py
```diff
@@ -75,11 +75,9 @@
...
-        (
-            maxtext_sweep_gke_test[i].run_with_run_name_generation()
-            >> maxtext_sweep_gke_test[
-                num_tests - 1 - i
-            ].run_with_run_name_generation()
-        )
+      prev = maxtext_sweep_gke_test[0].run_with_run_name_generation()
+      for i in range(1, len(maxtext_sweep_gke_test)):
+        curr = maxtext_sweep_gke_test[i].run_with_run_name_generation()
+        if i % 4 != 0:
```
RissyRan, line 81: Nit: let's move this 4 to a variable? i.e. something like `chain_num=4`?

RissyRan (approved):

dags/multipod/maxtext_v5e_configs_perf.py
raymondzouu: Done!

dags/multipod/maxtext_v5e_configs_perf.py
```diff
@@ -75,11 +75,10 @@
...
-                num_tests - 1 - i
-            ].run_with_run_name_generation()
-        )
+      chain_num = 4
+      prev = maxtext_sweep_gke_test[0].run_with_run_name_generation()
+      for i in range(1, len(maxtext_sweep_gke_test)):
+        curr = maxtext_sweep_gke_test[i].run_with_run_name_generation()
+        if i % chain_num != 0:
+          prev >> curr
+        prev = curr
```
jonb377, line 84: Not blocking, but I've been interested in trying out [chain](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/baseoperator/index.html#airflow.models.baseoperator.chain), it takes some number of tasks and adds dependencies within the groups (e.g. `chain([a, b], [c, d])` will chain a->b and c->d). The tricky part would be finding a clean way to break up the original list of tasks.

If you want to see how it looks, you can try something like this, but it's harder to read so I'd prefer to keep what you have:

```python
from airflow.models.baseoperator import chain
chain(*(maxtext_sweep_gke_test[n:n+chain_num] for n in range(0, len(maxtext_sweep_gke_test), chain_num)))
```

jonb377 (approved):
Thanks Raymond!

dags/multipod/maxtext_v5e_configs_perf.py
raymondzouu: Thanks for the suggestion! I'll give that a try in the future and keep it as is for now

PR #349, opened 2024-07-09
Return empty string if gcs file location not found

raymondzouu:
# Description

Return empty string if tensorboard file is not found in GCS. Before, an `AirflowFailException` was raised if the tensorboard file was missing causing the `post_process` task to fail. If the `post_process` task fails, any dependent tests will also fail. The tensorboard file usually goes missing when running the model fails and no tensorboard file is generated. In this case, the `post_process` task should still succeed and upload the pass/fail test status to bigquery. 

# Tests

Tested in local Composer env that post_process task still continues when tensorboard file in GCS is not found http://shortn/_3KUXYpZGGk

# Checklist

### DIFF
--- a/xlml/utils/metric.py
+++ b/xlml/utils/metric.py
@@ -266,6 +266,8 @@ def process_tensorboard_summary(
 
   if summary_config.use_regex_file_location:
     file_location = get_gcs_file_location_with_regex(file_location)
+    if file_location == "":
+      return [[]], [[]]
 
   aggregation_strategy = summary_config.aggregation_strategy
   include_tag_patterns = summary_config.include_tag_patterns
@@ -330,9 +332,8 @@ def get_gcs_file_location_with_regex(file_location: str) -> str:
         f"{next(filter(file_path_regex.match, all_blobs_names))}"
     )
   except StopIteration:
-    raise AirflowFailException(
-        f"No objects matched supplied regex: {file_location}"
-    )
+    logging.warning(f"No objects matched supplied regex: {file_location}")
+    return ""
 
 
 # TODO(qinwen): implement profile metrics & upload to Vertex AI TensorBoard


RissyRan:
I don't think we should overwrite the logic due to the situation `If the post_process task fails, any dependent tests will also fail....The tensorboard file usually goes missing when running the model fails and no tensorboard file is generated`.

Otherwise, this missing tensorboard file will be hidden. We should set up dependent tests properly instead. 

Any other thoughts?

raymondzouu:
> Otherwise, this missing tensorboard file will be hidden. We should set up dependent tests properly instead.

@RissyRan I'm not sure if I follow. If the tensorboard file is missing due to a failed workload, the `wait_for_workload_completion` will fail however the `post_process` should still succeed and upload the test status to bigquery. This will also allow any chained tests to continue to run.

xlml/utils/metric.py
```diff
@@ -330,9 +332,8 @@ def get_gcs_file_location_with_regex(file_location: str) -> str:
         f"{next(filter(file_path_regex.match, all_blobs_names))}"
     )
   except StopIteration:
-    raise AirflowFailException(
-        f"No objects matched supplied regex: {file_location}"
-    )
+    logging.info(f"No objects matched supplied regex: {file_location}")
```
RissyRan, line 335: Can we return a warning instead info?

xlml/utils/metric.py
```diff
@@ -266,6 +266,8 @@ def process_tensorboard_summary(
 
   if summary_config.use_regex_file_location:
     file_location = get_gcs_file_location_with_regex(file_location)
+    if file_location == "":
+      return [[]], [[]]
```
RissyRan, line 270: Can you double check if we return `[], []`? it seems this method return 2 lists.

xlml/utils/metric.py
raymondzouu: Done!

xlml/utils/metric.py
raymondzouu: I'm seeing this method returns (List[List[]], List[List[]]) https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/83da803d0aa587cf1c6f6b80ddc0aa39fa601f7f/xlml/utils/metric.py#L234-L242

RissyRan (approved):

PR #374, opened 2024-08-26
Jetstream-pytorch fix setup & add tests 

sixiang-google:
# Description

- Add quantization tests for llama-2 7b/13b/70b and llama3-8b
- Change request rate
- Fix a few setup issues


dags/inference/configs/jetstream_pytorch_gce_config.py
```diff
@@ -83,22 +83,26 @@ def get_jetstream_pytorch_inference_nightly_config(
       f"export CKPT_PATH={model_configs['checkpoint']}",
       f"export TOKENIZER_PATH=$(pwd)/ckpt_dir/{model_configs['tokenizer']}",
       f"export SHARDING_CONFIG={model_configs['sharding_config']}",
-      "mkdir ckpt_dir",
-      "gsutil cp -r ${CKPT_PATH}/* ckpt_dir/",
+      f"export QUANTIZE={str(model_configs['quantize'])}",
+      "mkdir /dev/shm/ckpt_dir",
```
yeandy, line 87: Curious, how is this different from `mkdir ckpt_dir`?

dags/inference/configs/jetstream_pytorch_gce_config.py
```diff
@@ -83,22 +83,26 @@ def get_jetstream_pytorch_inference_nightly_config(
...
       """python run_server.py \
         --model_name=${MODEL_NAME} \
         --size=${SIZE} \
         --batch_size=${BATCH_SIZE} \
         --max_cache_length=${MAX_CACHE_LEN} \
-        --checkpoint_path=ckpt_dir \
+        --checkpoint_path=/dev/shm/ckpt_dir \
         --tokenizer_path=${TOKENIZER_PATH} \
+        --quantize_weights=${QUANTIZE} \
+        --quantize_kv_cache=${QUANTIZE} \
```
yeandy, line 98: Since there may be cases where we don't quantize kv-cache, can we have a separate flag for these? i.e. 
```
        --quantize_weights=${QUANTIZE} \
        --quantize_kv_cache=${QUANTIZE_KV_CACHE} \
```

dags/inference/jetstream_pytorch_inference.py
```diff
@@ -42,12 +44,14 @@
...
           "sleep_time": 120,
           "tpu_version_cores": [(TpuVersion.V5E, 8)],
-          "checkpoint": "gs://inference-benchmarks/models/llama2-7b/pytorch/llama-2-7b-merged",
+          "checkpoint": "gs://inference-benchmarks/models/llama2-7b-chat/pytorch/llama-2-7b-chat-merged",
+          "checkpoint_quantized": "gs://inference-benchmarks/models/llama2-7b-chat/pytorch/llama-2-7b-chat-merged-int8-per-channel",
           "dataset": "openorca",
           "tokenizer": "tokenizer.llama2",
           "batch_sizes": [96],
-          "request_rate": 5,
+          "request_rate": 100,
```
yeandy, line 52: This is to essentially to mimic infinite rate, right?

dags/inference/jetstream_pytorch_inference.py
```diff
@@ -42,12 +44,14 @@
           "size": "7b",
           "sleep_time": 120,
           "tpu_version_cores": [(TpuVersion.V5E, 8)],
-          "checkpoint": "gs://inference-benchmarks/models/llama2-7b/pytorch/llama-2-7b-merged",
+          "checkpoint": "gs://inference-benchmarks/models/llama2-7b-chat/pytorch/llama-2-7b-chat-merged",
+          "checkpoint_quantized": "gs://inference-benchmarks/models/llama2-7b-chat/pytorch/llama-2-7b-chat-merged-int8-per-channel",
           "dataset": "openorca",
           "tokenizer": "tokenizer.llama2",
           "batch_sizes": [96],
```
yeandy, line 51: It might be useful to use a higher batch size for the quantized model. But in order to do this, we may need to create a separate entry in the `test_models` dictionary. i.e. 
```
"llama3-8b": {quantized: False, batch_sizes=[1, 32, 96]},
"llama3-8b-int8": {quantized: True, batch_sizes=[1, 96, 128, 160]},
```
though this may require a bit more parsing during the nested loop to extract the "llama3-8b" string from the `model` variable.

yeandy (commented):
Thanks @sixiang-google! can you please put successful DAG runs? Thanks!

dags/inference/configs/jetstream_pytorch_gce_config.py
sixiang-google: llama2-70b can't hold on the default disk

dags/inference/configs/jetstream_pytorch_gce_config.py
sixiang-google: Currently there's no such case in jetstream-pytorch, but I will add it

dags/inference/configs/jetstream_pytorch_gce_config.py
yeandy: Understood, thanks!

sixiang-google:
R: @FanhaiLu1

dags/inference/jetstream_pytorch_inference.py
sixiang-google: Currently we use 96 for both quantized and un-quantized models.

FanhaiLu1 (approved):
Thanks for the fix, looks good to me.

vipannalla (approved):
LGTM

PR #387, opened 2024-09-13
Move v4-128 tests to new cluster

bvandermoon:
# Description

This change moves all v4-128 tests to the new cluster. These tests have been failing since the previous cluster no longer exists.

# Tests

Ran one of the maxtext_convergence tests in local Airflow. The job is scheduled properly and runs on the new v4-128 cluster.

```
gcloud auth login --update-adc
bash scripts/local-airflow.sh dags/multipod/maxtext_convergence.py
```

Screenshot 2024-09-13 at 10 38 38 AM

Note that there is a failure at the process_metrics stage after the job has completed. This appears to be related to the local Airflow setup but let me know if I am missing something here.

# Checklist


dags/examples/xpk_example_dag.py
```diff
@@ -49,7 +49,7 @@
       tpu_cores=128,
       tpu_zone=Zone.US_CENTRAL2_B.value,
       test_name="resnet-multi-slice",
-      project_name=Project.TPU_PROD_ENV_MULTIPOD.value,
+      project_name=Project.CLOUD_TPU_MULTIPOD_DEV.value,
```
jonb377, line 52: It seems like we could make cluster changes like this much simpler by keeping all the cluster details together, like cluster name, device type, zone, etc. @RissyRan is there still a refactor planned for the codebase? Wondering if that's the right place for such improvements or if this would need to be its own work item.

jonb377 (approved):
Our test grid thanks you Branden!

dags/examples/xpk_example_dag.py
bvandermoon: +1

allenwang28 (approved):

dags/examples/xpk_example_dag.py
RissyRan: SGTM! Please feel free to refactor it.

PR #397, opened 2024-09-18
Centralize XPK cluster details

jonb377:
# Description

It is challenging and error-prone to update an XPK cluster. In https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/387, the cluster was renamed and moved regions, and subtle issues like applying the new zone to every test config were missed.

Some details are constant for each cluster, and we shouldn't need to pass them around in parallel. This change creates `XpkClusterConfig` to track the cluster config, and the single object can be used to define XPK tests.

# Tests

- Deployed to development env: https://fc137785b6c943a4af82d0671bf0ebe5-dot-us-central1.composer.googleusercontent.com/home
- Ran an example DAG: https://screenshot.googleplex.com/3dFCH5dZzWynuRb

# Checklist

### DIFF
--- a/dags/vm_resource.py
+++ b/dags/vm_resource.py
@@ -16,6 +16,7 @@
 
 import datetime
 import enum
+from xlml.apis.xpk_cluster_config import XpkClusterConfig
 
 
 V5_NETWORKS_PREFIX = "projects/tpu-prod-env-automated"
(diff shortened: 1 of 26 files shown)


jonb377:
cc @parambole @entrpn since this touches DAGs in the `imagegen_devx` folder.

dags/vm_resource.py
```diff
@@ -157,25 +158,86 @@ class RuntimeVersion(enum.Enum):
...
-  V5E_256_US_WEST_4_MULTISLICE_CLUSTER = "v5e-256-bodaborg-us-west4"
-  BODABORG_V6E_256 = "bodaborg-v6e-256"
-  A3_CLUSTER = "maxtext-a3-20n"
-  A3PLUS_CLUSTER = "a3plus-benchmark"
-  CPU_M1_MEGAMEM_96 = "m1-megamem-96-shared"
-  CPU_N2_STANDARD_64 = "shared-n2-standard-64"
+class XpkClusters:
+  """Common XPK cluster configs."""
+
+  V4_8_CLUSTER = XpkClusterConfig(
```
RissyRan, line 164: Thanks Jon! This is great! 

I see V4_8_CLUSTER and BODABORG_V6E_256. Wondering if it's better we just capitalize the cluster name itself as variable name? I am not sure how frequent we will update the cluster name and node numbers. Will it be possible single slice cluster become multislice in the future? or verse visa. 

I prefer one of following two options:
1) directly use cluster name here; or
2) format all of those in the format `<vm_type>_<vm_shape>_<optional identifier>_cluster`, for example TPU_V4_32_CLUSTER and GPU_A3_CLUSTER, CPU_M1_MEGAMEM_96_CLUSTER, etc. optional identifier could be single slice, multislice, etc.

How do you think?

RissyRan (commented):
Thanks Jon!

dags/vm_resource.py
jonb377: Synced up offline - we'll use the cluster name here.

dags/vm_resource.py
```diff
@@ -157,25 +158,86 @@ class RuntimeVersion(enum.Enum):
...
+      zone=Zone.US_CENTRAL2_B.value,
+  )
+  TPU_V4_16_CLUSTER = XpkClusterConfig(
+      name="v4-16-maxtext",
+      device_version=TpuVersion.V4,
+      core_count=16,
+      project=Project.TPU_PROD_ENV_MULTIPOD.value,
+      zone=Zone.US_CENTRAL2_B.value,
+  )
+  V4_128_MULTISLICE_CLUSTER = XpkClusterConfig(
```
jonb377, line 185: Oops, missed this one.

RissyRan (approved):
Thank you, Jon!

PR #420, opened 2024-09-26
Yijiaj/mlperf a2

jyj0w0:
# Description

Add MLPerf auto tuning for A2 and G2. 
Also add local_ssd option for GPU instance creating. 
Doc:https://cloud.google.com/compute/docs/disks/add-local-ssd#create_local_ssd

# Tests
https://screenshot.googleplex.com/5AXwwbUF7cqRuD8

# Checklist

- [ x] I have performed a self-review of my code.
- [ x] I have necessary comments in my code, particularly in hard-to-understand areas.
- [ x] I have run one-shot tests and provided workload links above if applicable. 
- [ x] I have made or will make corresponding changes to the doc if needed.


vipannalla (approved):

vipannalla:
@tohaowu, can you also review the PR?

xlml/utils/gpu.py
```diff
@@ -276,6 +299,7 @@ def wait_for_resource_creation(operation_name: airflow.XComArg):
                 "Error during resource creation: [Code:"
                 f" {operation.http_error_status_code}]:"
                 f" {operation.http_error_message}"
+                f" {operation.error}"
```
RissyRan, line 302: Do you have an example for this `operation.error`?

xlml/utils/gpu.py
```diff
@@ -168,7 +188,10 @@ def create_resource_request(
     machine_type = accelerator.machine_type
     image = get_image_from_family(project=image_project, family=image_family)
     disk_type = f"zones/{gcp.zone}/diskTypes/pd-ssd"
-    disks = [disk_from_image(disk_type, 100, True, image.self_link)]
+    disks = [disk_from_image(disk_type, 1000, True, image.self_link)]
```
RissyRan, line 191: Thanks for the change! Two questions:

1) is 1000GB needed for the case? it seems quite large, as needs by each accelerator.
2) Previously missed this hardcode `100` here. Could you help update method to below, so we could know it's a default value in helper function.

```
def disk_from_image(
    disk_type: str,
    boot: bool,
    source_image: str,
    disk_size_gb: int = 100,
    auto_delete: bool = True,
) -> compute_v1.AttachedDisk:
```

xlml/utils/gpu.py
jyj0w0: The error looks like this:  https://screenshot.googleplex.com/3cYYyU85shBeF3d 
I feel it might give more info though.

xlml/utils/gpu.py
jyj0w0: We will get "failed to register layer: write /usr/lib/x86_64-linux-gnu/libnvinfer_static.a: no space left on device" with only 100 gb so I make it as a optional input to GPU. Thanks!

xlml/utils/gpu.py
RissyRan: Thank you!

RissyRan (approved):
Thanks Yijia!

PR #433, opened 2024-10-14
Resubmit adding vllm benchmarks

richardsliu:
Adding vllm performance benchmarks to XLML tests. Supporting the following accelerator types initially:
* GPU (A100 40GB)
* TPU v5e-8 

Benchmark is using https://github.com/GoogleCloudPlatform/ai-on-gke/tree/main/benchmarks/benchmark/tools/profile-generator to generate requests.

# Checklist


xlml/utils/bigquery.py
```diff
@@ -112,6 +112,9 @@ def metadata_history_table_id(self):
 
   def is_valid_metric(self, value: float):
     """Check if float metric is valid for BigQuery table."""
+    if not isinstance(value, float):
```
RissyRan, line 115: I did not recall if I met any issue for this before. Probably not.

richardsliu: The new GKE benchmarks sometimes would have non-float metrics in some of the json output.

dags/solutions_team/solutionsteam_vllm_benchmarks.py
```diff
@@ -0,0 +1,127 @@
...
+from dags.vm_resource import GpuVersion, TpuVersion, Region, Zone, Project, INFERENCE_NETWORKS, H100_INFERENCE_SUBNETWORKS, V5_NETWORKS, V5E_SUBNETWORKS, V5P_SUBNETWORKS, ImageProject, ImageFamily, MachineVersion, RuntimeVersion
+from dags.multipod.configs.common import SetupMode, Platform
+from dags.solutions_team.configs.vllm import vllm_benchmark_config
+
+
+# Run once a day at 6 am UTC (10 pm PST)
+SCHEDULED_TIME = "0 6 * * *" if composer_env.is_prod_env() else None
+
+
+class AcceleratorType(enum.Enum):
```
RissyRan, line 17: Nit: could you move this to vm_resources.py? so that folks can reuse in the future. Please add CPU as well :)

RissyRan (commented):
LGTM! Thanks for the fix!

xlml/utils/bigquery.py
RissyRan: Are those values expected to be wiped out?

RissyRan:
If you could update the PR description for future reference, that will be great!

xlml/utils/bigquery.py
richardsliu: They are not used in our metrics, it's ok to discard them.

RissyRan: Will integers be filtered out?

```
>>> a = 1
>>> isinstance(1, float)
False
>>> isinstance(a, float)
False
```

RissyRan: Can you update it something like 

```
if isinstance(value, numbers.Number) and not (value, bool):
    invalid_values = [math.inf, -math.inf, math.nan]
    return not (value in invalid_values or math.isnan(value))

```

RissyRan (approved):
Thanks! Please feel free to merge once checks pass.

PR #447, opened 2024-10-24
Adding tests for M* workload built on top of JAX Stable Stack for GPUs

parambole:
# Description

Context: 

A) We have created a new JAX Stable Stack image for GPUs with JAX version 0.4.35.
B) We have added support for creating new M* docker images in https://github.com/AI-Hypercomputer/maxtext/pull/981.

In this PR, we are: 

1. Updating the JAX Stable Stack DAG to run Maxtext on GCP GPUs with images that are built using JAX Stable Stack.
2. mportant: Changing the notion of "stable" for Maxtext GPU images — i.e., from "Stable" to "stable stack," which now points to images built using JAX Stable Stack.

# Tests
image
image

# Checklist

### DIFF
--- a/dags/imagegen_devx/jax_stable_stack_e2e.py
+++ b/dags/imagegen_devx/jax_stable_stack_e2e.py
@@ -51,7 +51,10 @@ with models.DAG(
       "steps=2 enable_checkpointing=false attention=dot_product"
   )
   test_models_gpu = {
-      "train-c4-data": (f"{train_base} run_name=runner-{current_datetime}-0", 1),
+      "train-c4-data": (
+          f"{train_base} run_name=runner-{current_datetime}-0",
+          1,
+      ),
   }
 
   for accelerator, slices in maxtext_test_configs.items():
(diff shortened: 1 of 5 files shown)


RissyRan (approved):
LGTM! Please get @jonb377's approval for multipod's tests. Thank you!

jonb377 (approved):
Thanks Param!

dags/imagegen_devx/jax_stable_stack_e2e.py
```diff
@@ -43,12 +44,24 @@
       "v4-8": [1],
       "v6e-256": [1],
   }
-  current_datetime = config.get_current_datetime()
+  train_base = (
+      "XLA_PYTHON_CLIENT_MEM_FRACTION=0.65 TF_FORCE_GPU_ALLOW_GROWTH=true "
```
bvandermoon, line 48: Just wondering what is the context for lowering XLA_PYTHON_CLIENT_MEM_FRACTION to 0.65 here?

bvandermoon (approved):
LGTM, thanks @parambole

dags/imagegen_devx/jax_stable_stack_e2e.py
parambole: @bvandermoon I was using the same train command as in https://github.com/GoogleCloudPlatform/ml-auto-solutions/blame/master/dags/multipod/maxtext_gpu_end_to_end.py#L40.

As far as I understand, `XLA_PYTHON_CLIENT_MEM_FRACTION` specifies how much memory JAX<>XLA would have available. I am unsure what the reasoning for the original train command was, i.e., if extra memory was being reserved for operations other than JAX. But thinking about it, I believe this parameter is not required here for the tests in any case, so I have removed it.

PR #459, opened 2024-11-01
Add tests for Jax.distributed.initialize with v4 and v5p TPUs

guptaaka:
# Description

Add tests for Jax.distributed.initialize with v4 and v5p TPUs.

# Tests

Tested the changes locally using the dev environment.

bash scripts/upload-tests.sh gs://us-central1-ml-automation-s-bc6c8818-bucket/dags

http://shortn/_BosNVvrqI1

# Checklist

### DIFF
--- a/dags/multipod/configs/jax_tests_gce_config.py
+++ b/dags/multipod/configs/jax_tests_gce_config.py
@@ -22,8 +22,7 @@ import datetime
 
 PROJECT_NAME = Project.CLOUD_ML_AUTO_SOLUTIONS.value
 RUNTIME_IMAGE = RuntimeVersion.TPU_UBUNTU2204_BASE.value
-# PROJECT_NAME = Project.TPU_PROD_ENV_MULTIPOD.value
-# RUNTIME_IMAGE = RuntimeVersion.TPU_UBUNTU2204_BASE.value
+
 
 def get_jax_distributed_initialize_config(
     tpu_version: TpuVersion,
@@ -37,19 +36,17 @@ def get_jax_distributed_initialize_config(
     network: str = "default",
     subnetwork: str = "default",
     is_tpu_reserved: bool = True,
-    automated_test: bool = True,
     num_slices: int = 1,
-    base_output_dir: str = gcs_bucket.BASE_OUTPUT_DIR,
 ):
-  set_up_cmds = (
-    'pip install "jax[tpu]" -f https://storage.googleapis.com/jax-releases/libtpu_releases.html,
-    'python -m pip install fabric;',
-  )
-  run_model_cmds = (
-      (
-          "python -m 'print(10111); import jax; jax.distributed.initialize(); print(20111);'"
-      ),
-  ) # TODO: Do we need to verify success or will the command fail if execution fails?
+  test_platform = common.Platform.GCE
+  set_up_cmds = common.setup_maxtext(test_mode, test_platform)
+  set_up_cmds = [
+    "pip install 'jax[tpu]' -f https://storage.googleapis.com/jax-releases/libtpu_releases.html",
+  ]
+  run_model_cmds = [
+    "python3 -c 'import jax; jax.distributed.initialize()'",
+  ]
+
   job_test_config = test_config.TpuVmTest(
       test_config.Tpu(
           version=tpu_version,
--- a/dags/multipod/configs/jax_tests_gce_config.py
+++ b/dags/multipod/configs/jax_tests_gce_config.py
@@ -41,7 +41,7 @@ def get_jax_distributed_initialize_config(
     num_slices: int = 1,
     base_output_dir: str = gcs_bucket.BASE_OUTPUT_DIR,
 ):
-  set_up_cmds = 'pip install "jax[tpu]" -f https://storage.googleapis.com/jax-releases/libtpu_releases.html; python -m pip install fabric;'
+  set_up_cmds = ('pip install "jax[tpu]" -f https://storage.googleapis.com/jax-releases/libtpu_releases.html; python -m pip install fabric;')
   run_model_cmds = (
       (
           "python -m 'print(10111); import jax; jax.distributed.initialize(); print(20111);'"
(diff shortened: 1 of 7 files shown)


dags/multipod/configs/jax_tests_gce_config.py
```diff
@@ -0,0 +1,76 @@
...
+    runtime_version: str = RUNTIME_IMAGE,
+    network: str = "default",
+    subnetwork: str = "default",
+    is_tpu_reserved: bool = True,
+    num_slices: int = 1,
+):
+  test_platform = common.Platform.GCE
+  set_up_cmds = common.setup_maxtext(test_mode, test_platform)
+  set_up_cmds = [
+      "pip install 'jax[tpu]' -f https://storage.googleapis.com/jax-releases/libtpu_releases.html",
```
gobbleturk, line 44: I can think of three possible jax versions to test
1) latest stable (this PR)
2) nightly
3) latext jax stable

There is some value in testing all 3, but just testing one latest stable here already gives a lot of value as well

deployment/cloud_composer.auto.tfvars
```diff
@@ -13,4 +13,4 @@ environment_config = [
     environment_name   = "ml-automation-solutions-dev"
     service_account_id = "ml-auto-solutions-dev"
   }
-]
```
gobbleturk, line 16: nit: clean up whitespace in untouched files

gobbleturk (approved):

dags/multipod/configs/jax_tests_gce_config.py
guptaaka: Good point, I'll add the other 2 in a follow-on PR. Let's first get this test running.

deployment/cloud_composer.auto.tfvars
guptaaka: I'm actually not sure why this is showing up as an edit. I'd simply ignore this change, as the file automatically adds an empty line at the end.

dags/multipod/configs/jax_tests_gce_config.py
```diff
@@ -0,0 +1,76 @@
+# Copyright 2023 Google LLC
```
RissyRan, line 1: Nit: 2024

dags/multipod/jax_functional_tests.py
```diff
@@ -0,0 +1,100 @@
...
+          tpu_cores=8,
+          tpu_zone=Zone.US_CENTRAL2_B.value,
+          time_out_in_min=60,
+          is_tpu_reserved=False,
+          test_name=default_test_name,
+          test_mode=test_mode,
+      )
+  )
+
+  jax_nightly_2slice_v4_8 = (
```
RissyRan, line 50: Do you want to chain those tests up? This gives us more flexibility (add this at the end, similar to v5p):

`jax_nightly_1slice_v4_8 >> jax_nightly_2slice_v4_8`

1) when you adding more tests in the future, your DAG will not trigger whole tests at one time.

2) when a test in 1 slice, and most likely it will fail at the same reason in 2 slices. Due to the current setting (clean_up step will always be successful, no test after that will be skipped), it will be a little bit hard to skip the 2 slice, but we could keep this logic in mind.

RissyRan (commented):
Thanks! I'd like to suggest to chain those tests together (v4_1slice >> v4_2slice and v5_1slice >> v5_2slice), so they won't use all test capacity at once when adding more tests in the future.

guptaaka:
> Thanks! I'd like to suggest to chain those tests together (v4_1slice >> v4_2slice and v5_1slice >> v5_2slice), so they won't use all test capacity at once when adding more tests in the future.

Good point. I'm using very small slices right now, I think it should be okay to trigger the tests altogether. We can chain these as we scale up.

dags/multipod/configs/jax_tests_gce_config.py
guptaaka: Good catch

RissyRan:
> > Thanks! I'd like to suggest to chain those tests together (v4_1slice >> v4_2slice and v5_1slice >> v5_2slice), so they won't use all test capacity at once when adding more tests in the future.
> 
> Good point. I'm using very small slices right now, I think it should be okay to trigger the tests altogether. We can chain these as we scale up.

I think it's better to chain them up now, unless we have a strong reason do not. Sometimes we just forget in future PRs. We have lots of DAGs have chained them up, and just a few of them:

1) https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/dags/multipod/maxtext_end_to_end.py#L175
2) https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/dags/multipod/maxtext_gpu_end_to_end.py#L146
3) https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/dags/multipod/mxla_collective_nightly.py#L91-L95

dags/multipod/jax_functional_tests.py
guptaaka: Done! Thanks

RissyRan (approved):
Thank you!

dags/multipod/configs/jax_tests_gce_config.py
```diff
@@ -0,0 +1,76 @@
...
+  test_platform = common.Platform.GCE
+  set_up_cmds = common.setup_maxtext(test_mode, test_platform)
+  set_up_cmds = [
+      "pip install 'jax[tpu]' -f https://storage.googleapis.com/jax-releases/libtpu_releases.html",
+  ]
+  run_model_cmds = [
+      "python3 -c 'import jax; jax.distributed.initialize()'",
+  ]
+
+  job_test_config = test_config.TpuVmTest(
```
jonb377, line 50: Any reason to use CQR over XPK?

jonb377 (approved):
LGTM

dags/multipod/configs/jax_tests_gce_config.py
guptaaka: We want to test both GCE and GKE stacks. I'll add the GKE tests after this PR.

PR #496, opened 2024-12-04
Remove TF SE tests

wenxindongwork:
These StreamExecutor tests are no longer maintained and will be removed.


dags/solutions_team/solutionsteam_tf_nightly_supported.py
```diff
@@ -112,7 +112,7 @@
       runtime_version=RuntimeVersion.TPU_VM_TF_NIGHTLY_POD.value,
   )
 
-  embedding_dim = 32
```
richardsliu, line 115: is this intended to be part of this PR?

chandrasekhard2 (approved):

dags/solutions_team/solutionsteam_tf_nightly_supported.py
wenxindongwork: this was from upstream, I just merged with master so this change is now gone.

PR #499, opened 2024-12-05
Update the tpu dependency installation command

tengyifei:
# Description

Since libtpu is switching Python package registries, we need to include both registries during the transition. https://github.com/pytorch/xla/blob/master/README.md is already updated but we need to update these hardcoded commands here as well.

# Tests

Verified things in a test composer instance:

```
bash scripts/upload-tests.sh gs://us-central1-piz-a9563b85-bucket/dags
# Check results in http://shortn/_xCUuG2UI33
```

# Checklist


ManfeiBai (approved):
Thanks, let's megre after we saw green run in http://shortn/_xCUuG2UI33

PR #538, opened 2025-01-10
Reuse Exist GPU Instance for Auotmation

jyj0w0:
# Description

Add functions for reusing GPU instance. 

# Tests

https://screenshot.googleplex.com/3uLrquTpusRR2eH

# Checklist


xlml/utils/gpu.py
```diff
@@ -129,6 +129,128 @@ def generate_gpu_name() -> str:
...
+      item.value = (
+          item.value + "\n" + f"cloud-ml-auto-solutions:{ssh_keys.public}"
+      )
+      break
+  if not ssh_key_exist:
+    items.append({
+        "key": "ssh-keys",
+        "value": f"cloud-ml-auto-solutions:{ssh_keys.public}",
+    })
+    metadata.items = items
```
zpcore, line 175: Maybe we don't need this assignment because is `items` and `metadata.items` since they are alias.

xlml/utils/gpu.py
```diff
@@ -129,6 +129,128 @@ def generate_gpu_name() -> str:
...
+      zone=gcp.zone,
+  )
+  instance = instance_client.get(request=instance_request)
+  logging.info(
+      f"Resource get status: {instance.status}, {instance.status_message}"
+  )
+  metadata = instance.metadata
+  for item in metadata.items:
+    if item.key == "ssh-keys":
+      item.value = item.value.replace(
```
zpcore, line 227: Can you elaborate more why the `ssh-keys` got replaced here?

xlml/utils/gpu.py
```diff
@@ -129,6 +129,128 @@ def generate_gpu_name() -> str:
...
+
+  ip_address = instance.network_interfaces[0].network_i_p
+  metadata = instance.metadata
+  items = metadata.items or []
+  ssh_key_exist = False
+  for item in metadata.items:
+    if item.key == "ssh-keys":
+      ssh_key_exist = True
+      item.value = (
+          item.value + "\n" + f"cloud-ml-auto-solutions:{ssh_keys.public}"
```
zpcore, line 167: Looks like you are overwriting the `ssh-keys` if the key exists. I assume all the existing key will be overwritten, is this true?

zpcore (commented):
LGTM, left a few question.

xlml/utils/gpu.py
jyj0w0: Not really, it appends new one to the already existed ones. In portal I can see existing ones are not affected.

xlml/utils/gpu.py
jyj0w0: I think the metadata.items might not always exist so assigning it in case we create a new list.

xlml/utils/gpu.py
jyj0w0: This replace the newly added ssh into an empty string so that the ssh setting could be same as DAG runs.

xlml/utils/gpu.py
zpcore: I see, thanks!

xlml/utils/gpu.py
zpcore: Hmm, I didn't get that. Do you mean ssh setting can be changed after the provision stage if we don't replace it here?

xlml/apis/task.py
```diff
@@ -349,6 +349,7 @@ class GpuCreateResourceTask(BaseTask):
     task_metric_config: metric configuration (e.g., result gcs path).
     gpu_create_timeout: timeout when waiting for the GPU vm creation.
     install_nvidia_drivers: whether to install Nvidia drivers.
+    existed_instance_name: whether a exited GPU instance shall be used.
```
richardsliu, line 352: s/exited/existing

dags/inference/configs/trt_llm_inference_config.py
```diff
@@ -36,6 +36,7 @@ def get_trt_llm_gpu_config(
     project: Project,
     network: str,
     subnetwork: str,
+    existed_instance_name: str = None,
```
richardsliu, line 39: Existed means the instance doesn't exist now. Maybe change to "existing"?

xlml/apis/task.py
```diff
@@ -399,6 +405,58 @@ def run(self) -> DAGNode:
       provision >> run_model >> post_process >> clean_up
     return group
 
+  def run_with_existed_instance(self) -> DAGNode:
+    """Run a test job.
+
+    Returns:
+      A task group with the following tasks chained:  run_model and post_process.
```
RissyRan, line 412: Nit: should we still include provision and clean_up, based on `provision >> run_model >> post_process >> clean_up`?

xlml/utils/metric.py
```diff
@@ -631,12 +637,15 @@ def get_gce_job_status(
       return bigquery.JobStatus.MISSED
 
     # check setup status to see if setup step is successful
-    setup_task = current_dag.get_task(task_id=f"{benchmark_id}.provision.setup")
-    setup_ti = TaskInstance(setup_task, execution_date)
-    setup_state = setup_ti.current_state()
-    if setup_state == TaskState.FAILED.value:
-      logging.info("The setup state is failed, and the job status is failed.")
-      return bigquery.JobStatus.FAILED
+    if not use_existed_instance:
```
RissyRan, line 640: should we add case of `use_existed_instance=True`?

xlml/apis/task.py
```diff
@@ -491,6 +551,7 @@ def post_process(
           self.task_metric_config,
           self.task_gcp_config,
           folder_location=result_location,
+          use_existed_instance=use_existed_instance,
```
RissyRan, line 554: Should we wrap `use_existed_instance` into [test_config](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/test_config.py#L148) or its dependent class? I recall the the reason add folder_location is related to environment variables, and we don't have a way to workaround.

RissyRan (commented):
Thanks for adding the feature!

xlml/utils/gpu.py
zpcore: I see, this is the last stage of the run. Thanks for the detail!

xlml/apis/task.py
jyj0w0: Probably add to the [GpuVmTest[(https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/apis/test_config.py#L220)? Since I am only editing to GPU related features.

dags/inference/configs/trt_llm_inference_config.py
jyj0w0: Editted

xlml/apis/task.py
jyj0w0: Included.

xlml/utils/metric.py
jyj0w0: Added.

xlml/apis/task.py
jyj0w0: Oh I missed that this parameter is a `TestConfig` not a `GpuVmTest`.  Currently I think only for `GpuVmTest` it is tested, not sure if the parameter should go to the upper level `TestConfig` though.

xlml/apis/task.py
```diff
@@ -349,6 +349,7 @@ class GpuCreateResourceTask(BaseTask):
     task_metric_config: metric configuration (e.g., result gcs path).
     gpu_create_timeout: timeout when waiting for the GPU vm creation.
     install_nvidia_drivers: whether to install Nvidia drivers.
+    existing_instance_name: whether a existing GPU instance shall be used.
```
RissyRan, line 352: nit: `an existing...`

xlml/apis/task.py
RissyRan: I think you could add it into GpuVmTest with default value, one example: `num_slices` filed in TPU, and how to get it: https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/utils/metric.py#L447. Unfortunately, it will need a `if` check there.

RissyRan (commented):
Others LGTM! Thanks!

xlml/apis/task.py
jyj0w0: Updated, thanks!

richardsliu (approved):

vipannalla (approved):

ManfeiBai (approved):
Thanks, LGTM

PR #545, opened 2025-01-14
Add mlscale_onduty tag to more Dags

jiangjy1982:
# Description
As title.

# Tests
N/A

# Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ webserver_config.py
 logs
 
 # Misc
+swp
 tmp
 
 # Build artifacts
--- a/dags/multipod/maxtext_checkpointing.py
+++ b/dags/multipod/maxtext_checkpointing.py
@@ -29,7 +29,7 @@ SCHEDULED_TIME = "0 10 * * *" if composer_env.is_prod_env() else None
 with models.DAG(
     dag_id="maxtext_checkpointing",
     schedule=SCHEDULED_TIME,
-    tags=["multipod_team", "maxtext", "stable", "nightly"],
+    tags=["multipod_team", "mlscale_onduty", "maxtext", "stable", "nightly"],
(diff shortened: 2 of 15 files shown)


.gitignore
```diff
@@ -20,6 +20,7 @@ webserver_config.py
 logs
 
 # Misc
+swp
```
shralex, line 23: what does this do ?

shralex (approved):
Please consider also removing our tag from maxtext_sweep_gke_example_dag and pytorch_multislice

.gitignore
jiangjy1982: It's swp file generated by vim (Yes, I'm using Vim :)

jiangjy1982: Also, I found a new tag "mlscale_onduty" introduced in https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/501. I think it's a better idea to use the new tag for onduty purpose and keep the "multipod_team" tag  to indicate relevancy. So as a result I ended up adding this tag to a bunch of DAGs (and remove it from some).

shralex: Sounds good!

RissyRan (approved):
Thanks Jet! LGTM!

I also see multipod_team tag, how should we distinguish it with `mlscale_onduty`. If they serve the same purpose (owned by ML@scale), shall we remove it? Not necessary in this PR.

jiangjy1982:
> Thanks Jet! LGTM!
> 
> I also see multipod_team tag, how should we distinguish it with `mlscale_onduty`. If they serve the same purpose (owned by ML@scale), shall we remove it? Not necessary in this PR.

I kept the 'multipod_team' tag because it's serving a different semantic than 'mlscale_onduty' - it simply means the DAG is somewhat related to the multipod team, but not necessarily monitored by the onduty.

PR #556, opened 2025-01-15
Update XLML tests for MaxText on GPUs. 

yangyuwei:
# Description

There are two major changes:
1. Update the tests to scale up the A3 cluster and scale down the cluster before and after the tests run, to resolve the issue that the cluster gets automatically scaled down while it's not being used.
2. Replace the owner of the tests with Yuwei.

# Tests

1. Run the following commands:
$ git clone -b yangyuwei-maxtext-gpu https://github.com/GoogleCloudPlatform/ml-a
uto-solutions.git
$ cd ml-auto-solutions/
$ bash scripts/upload-tests.sh  gs://us-central1-nina-dags-fe347643-bucket/dags
2. Manually trigger `maxtext_gpu_end_to_end` DAG on the UI: https://540cef12d3da42ce97b48a97401b6c83-dot-us-central1.composer.googleusercontent.com/home

**List links for your tests (use go/shortn-gen for any internal link):** https://540cef12d3da42ce97b48a97401b6c83-dot-us-central1.composer.googleusercontent.com/dags/maxtext_gpu_end_to_end/grid?task_id=scale_down.scale_down_a3_cluster

# Checklist

- [ Y ] I have performed a self-review of my code.
- [ Y ] I have necessary comments in my code, particularly in hard-to-understand areas.
- [ Y ] I have run one-shot tests and provided workload links above if applicable. 
- [ Y ] I have made or will make corresponding changes to the doc if needed.

### DIFF
--- a/dags/multipod/maxtext_gpu_end_to_end.py
+++ b/dags/multipod/maxtext_gpu_end_to_end.py
@@ -253,5 +253,3 @@ with models.DAG(
     scale_down_a3_cluster()
 
   scale_up >> run_tests >> scale_down
-
-
--- a/dags/multipod/maxtext_gpu_end_to_end.py
+++ b/dags/multipod/maxtext_gpu_end_to_end.py
@@ -32,7 +32,7 @@ from xlml.utils import gke
 SCHEDULED_TIME = "0 4 * * *" if composer_env.is_prod_env() else None
 
 # Number of nodes on A3 cluster to be scaled up to
-A3_NUM_NODES = 16
+A3_NUM_NODES = 10
 
 
 def configure_project_and_cluster(project: str, cluster_name: str, zone: str):
--- a/dags/multipod/maxtext_gpu_end_to_end.py
+++ b/dags/multipod/maxtext_gpu_end_to_end.py
@@ -31,6 +31,9 @@ from xlml.utils import gke
 # Run once a day at 4 am UTC (8 pm PST)
 SCHEDULED_TIME = "0 4 * * *" if composer_env.is_prod_env() else None
 
+# Number of nodes on A3 cluster to be scaled up to
+A3_NUM_NODES = 8
+
 
 def configure_project_and_cluster(project: str, cluster_name: str, zone: str):
   region = gke.zone_to_region(zone)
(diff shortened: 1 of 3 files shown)


shralex (approved):
Looks great, thanks Yuwei!

dags/multipod/maxtext_gpu_end_to_end.py
```diff
@@ -146,6 +229,27 @@
...
+    scale_up_a3_cluster()
+
+  with TaskGroup(
+      group_id="run_tests", dag=dag, prefix_group_id=False
+  ) as run_tests:
+    run_maxtext_tests(dag)
+
+  with TaskGroup(group_id="scale_down", dag=dag) as scale_down:
+    scale_down_a3_cluster()
+
```
parambole, line 254: @yangyuwei, is my understanding correct?

1. The cluster initially has zero nodes.
2. We scale it up to A3_NUM_NODES nodes.
3. We scale it back down to zero.

Questions:

1. Is this cluster now solely to be used by the maxtext_gpu_end_to_end.py DAG?
2. We have also been using this cluster in https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/dags/sparsity_diffusion_devx/jax_stable_stack_gpu_e2e.py. I assume we would require similar logic for scaling the cluster up and down in that DAG too? This might present a race condition while scaling the cluster.

Is there a way to keep the cluster running at 10 nodes permanently?

yangyuwei: Hi @parambole that's a very good question. There is no way to keep the cluster running at 10 nodes permanently at the moment. Without this PR, the cluster can be scaled down at any time without notice due to the fact that it's not being used, and that's what happened in the past days.

To resolve the issue you mentioned in #2, I think there might be two options:
1. Merge maxtext_gpu_end_to_end.py and jax_stable_stack_gpu_e2e.py as one DAG
2. Add similar logic for jax_stable_stack_gpu_e2e.py to create another cluster, which should avoid the race condition you mentioned (I can reduce `A3_NUM_NODES` so that we can share the capacity)

What do you think?

parambole (approved):

PR #561, opened 2025-01-16
force llama2 spmd train on v5p use version `v2-alpha-tpuv5`

ManfeiBai:
# Description

force llama2 spmd train on v5p use version `v2-alpha-tpuv5`

# Tests

Airflow UI: https://dade15488e304a7dadc2e167f18721ee-dot-us-central1.composer.googleusercontent.com/dags/pytorchxla-r2-6/grid?dag_run_id=manual__2025-01-16T01%3A21%3A19.300605%2B00%3A00&task_id=pt-2-6-llama2-train-spmd-func-v5p-8-1vm.run_model

# Checklist

### DIFF
--- a/dags/legacy_test/tests/pytorch/r2.6/llama2-model.libsonnet
+++ b/dags/legacy_test/tests/pytorch/r2.6/llama2-model.libsonnet
@@ -122,6 +122,11 @@ local utils = import 'templates/utils.libsonnet';
         sudo pip3 install -e . -c ~/hf-constraints.txt
         pip3 install datasets evaluate scikit-learn accelerate -c ~/hf-constraints.txt
 
+        echo "run uninstall libtpu"
+        pip uninstall libtpu -y
+        pip install https://storage.googleapis.com/libtpu-nightly-releases/wheels/libtpu-nightly/libtpu_nightly-0.1.dev20250101+nightly-py3-none-linux_x86_64.whl
+        pip list | grep libtpu
+
         cd
         # 7B config
         mkdir 7B
--- a/dags/legacy_test/tests/pytorch/r2.6/llama2-model.libsonnet
+++ b/dags/legacy_test/tests/pytorch/r2.6/llama2-model.libsonnet
@@ -122,11 +122,6 @@ local utils = import 'templates/utils.libsonnet';
         sudo pip3 install -e . -c ~/hf-constraints.txt
         pip3 install datasets evaluate scikit-learn accelerate -c ~/hf-constraints.txt
 
-        echo "run uninstall libtpu"
-        pip uninstall libtpu -y
-        pip install https://storage.googleapis.com/libtpu-nightly-releases/wheels/libtpu-nightly/libtpu_nightly-0.1.dev20250101+nightly-py3-none-linux_x86_64.whl
-        pip list | grep libtpu
-
         cd
         # 7B config
         mkdir 7B
@@ -212,7 +207,7 @@ local utils = import 'templates/utils.libsonnet';
     llama2 + v4_8 + infer + timeouts.Hours(3),
     llama2 + v4_8 + spmd + timeouts.Hours(3),
     llama2 + v5p_8 + infer + timeouts.Hours(3),
-    llama2 + v5p_8 + spmd + timeouts.Hours(3),
+    llama2 + spmd + v5p_8 + timeouts.Hours(3),
     llama3_train + v5p_8 + timeouts.Hours(3),
     llama3_train + trillium_4 + timeouts.Hours(3),
   ],


bhavya01:
Discussed offline that the release branch uses libtpu stable version 0.0.6 and we should test with that version. The actual fix would to change the runtime_version while creating the TPUv5 from `ubuntu2204` to `v2-alpha-tpuv5`

tengyifei (changes requested):
see bhavya's comment

ManfeiBai:
> Discussed offline that the release branch uses libtpu stable version 0.0.6 and we should test with that version. The actual fix would to change the runtime_version while creating the TPUv5 from `ubuntu2204` to `v2-alpha-tpuv5`

tested when only change version to `v2-alpha-tpuv5` still met the error: http://shortn/_s8p3EXJSLR

tried to update libtpu version to 0.0.7 also fixed the issue: https://gist.github.com/ManfeiBai/e695a83b602b80d09427418327fe4a3b

will only keep version change in current PR, and rebuild a new rc wheel with libtpu-0.0.7

ManfeiBai:
> see bhavya's comment

thanks, replied

bhavya01 (approved):

bhavya01:
Can we update the PR description too? I will leave the decision to update the libtpu to Yifei

tengyifei:
> decision to update libtpu 

Seems worth a try to me!

ManfeiBai:
> Can we update the PR description too? I will leave the decision to update the libtpu to Yifei

thanks, updated the description to only include version change;

created PR for libtpu in https://github.com/pytorch/xla/pull/8585 too

ManfeiBai:
> see bhavya's comment

Hi, @tengyifei, would you mind help to confirm the request change?

tengyifei:
@ManfeiBai not sure I understand -- this PR seems a no-op to me. It reorders two parameters in a jsonnet. 

IIUC we'd try bumping libtpu and Jax (and maybe openxla) in the 2.6 branch of torch_xla?

ManfeiBai:
> @ManfeiBai not sure I understand -- this PR seems a no-op to me. It reorders two parameters in a jsonnet.
> 
> IIUC we'd try bumping libtpu and Jax (and maybe openxla) in the 2.6 branch of torch_xla?

Thanks for the reply!

Current solution is:
1. update version to `v2-alpha-tpuv5`:
    1.1 changed in current PR according to @bhavya01 suggestion:
        1.1.1 `spmd` generated from `PyTorchTpuVmMixin`: [link](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/cfed7ed928dd7df286452edaca43fa595aa09a90/dags/legacy_test/tests/pytorch/r2.6/llama2-model.libsonnet#L72C37-L72C54)
        1.1.2 `PyTorchTpuVmMixin` was set with version `tpu-ubuntu2204-base`: [link](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/b7457b2c36beefecd55b343986f120b88078f95c/dags/legacy_test/tests/pytorch/r2.6/common.libsonnet#L92)
        1.1.3 current order might cause `tpu-ubuntu2204-base` overwrite `v2-alpha-tpuv5` defined by [`v5p_8`](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/cfed7ed928dd7df286452edaca43fa595aa09a90/dags/legacy_test/tests/pytorch/r2.6/llama2-model.libsonnet#L193)(which is defined with `v2-alpha-tpuv5`), so propose to switch the order of `spmd` and `v5p_8` to make sure version used is `v2-alpha-tpuv5`
        1.1.4 to confirm whether this order change affect test result, we need to block current PR, and wait for Airflow UI test result after below two PRs merge and rc7 wheel been built
2. bumping libtpu and Jax (and maybe openxla) in the 2.6 branch of torch_xla:
    2.1 https://github.com/pytorch/xla/pull/8585
    2.2 https://github.com/pytorch/xla/pull/8586
    
 will block this PR until rc7 wheel is built

tengyifei:
I see. IIUC the v5p-alpha image thing should always be used to run code with TPUv5p chips. I'd suggest checking the system image version via other means (for example, you could maybe run this DAG in a staging airflow instance, and then go to Pantheon to check the system image of the TPU VM). If the system image is wrong (ubuntu base), then we should still fix that e.g. by swapping the order of these jsonnet mixing irregardless of the test result.

tengyifei (approved):
I'll approve this change but see my comments

tengyifei:
I think I still don't understand what's going on because:
- If this test was failing because the VM image is wrong, then it should be fixed by switching to the correct VM image, and we don't need a libtpu pin update 
If this test was failing because of a libtpu bug, then it should be fixed by bumping libtpu

Unless.... both the VM image is wrong and also there's a libtpu bug?

ManfeiBai:
> both the VM image is wrong and also there's a libtpu bug?

I would agree with ` both the VM image is wrong and also there's a libtpu bug`, since local passed test with newer version and newer libtpu,

but I also doubt the VM imager version is one of the root cause too, we will only be able to confirm that after rc7 is built

PR #562, opened 2025-01-16
rename run_workload with name

ManfeiBai:
# Description

rename run_workload with name

# Test
image

# Checklist

### DIFF
--- a/dags/mantaray/run_mantaray_jobs.py
+++ b/dags/mantaray/run_mantaray_jobs.py
@@ -47,7 +47,7 @@ if composer_env.is_prod_env() or composer_env.is_dev_env():
       catchup=False,
   ) as dag:
     for workload_file_name in workload_file_name_list:
-      run_workload = mantaray.run_workload.override(task_id=workload_file_name)(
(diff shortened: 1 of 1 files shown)


wenxindongwork (approved):

dags/mantaray/run_mantaray_jobs.py
```diff
@@ -47,7 +47,9 @@
       catchup=False,
   ) as dag:
     for workload_file_name in workload_file_name_list:
-      run_workload = mantaray.run_workload(
+      run_workload = mantaray.run_workload.override(
+          task_id=workload_file_name[:-3]
```
tengyifei, line 51: nit: I saw one task with a "." at the end. Is that because the workload file name extension may not be 3 characters? Is it possible to use `split` to split out the extension cleanly?

tengyifei (approved):

dags/mantaray/run_mantaray_jobs.py
ManfeiBai: Good suggestion, updated to `split` solution

PR #566, opened 2025-01-17
Maxtext GPU test now point to image built using mode=stable

parambole:
# Description

Since JStS image is not yet publicly released reverting changes so that maxtext gpu test now point to image which is built using `mode=stable` 

ref: https://github.com/AI-Hypercomputer/maxtext/pull/1173 

# Checklist


shralex (approved):

yangyuwei (commented):
Thanks Param for the quick fix! 

I think it's better to hold on merging this PR until https://github.com/AI-Hypercomputer/maxtext/pull/1173 is merged and executed, otherwise, there will be no `stable` docker images that can be used by XLML tests. 

In addition, it'd be great if you could run the tests to verify after the new `stable` docker image gets uploaded.

Does it make sense?

PR #575, opened 2025-01-23
Pathways job completion fix

lukebaumann:
# Description

Pathways jobs delete all pods on failure so the current `wait_for_workload_completion` does not recognize completion correctly. The new completion logic should be better than the current "check all pods" logic to determine if a workload is complete but because other logic with GPU workloads and print logs depending on the pods is colocated, I made only minimally invasive changes.

# Tests

Ran both the maxtext_v5e_config_perf and pathways_maxtext_v5e_config_perf DAGs and they are WAI.

# Checklist

### DIFF
--- a/xlml/utils/xpk.py
+++ b/xlml/utils/xpk.py
@@ -144,6 +144,42 @@ def _list_workload_pods(
   return pods
 
 
+def _get_batch_api_client(
+    project_id: str, region: str, cluster_name: str
+) -> k8s_client.BatchV1Api:
+  """Create a batch API client for the given cluster."""
+  client = gke.get_authenticated_client(project_id, region, cluster_name)
+
+  # Initilize the client
+  batch_api = k8s_client.BatchV1Api(client)
+  logging.info(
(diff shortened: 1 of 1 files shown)


RissyRan (commented):
LGTM! Have you tested in XL ML airflow that won't present any unexpected behavior?

lukebaumann:
Ran both the maxtext_v5e_config_perf and pathways_maxtext_v5e_config_perf DAGs and they are WAI.

There is an existing pathways bug that is causing the DAG to fail and the failure logic is WAI.

I am not sure of a couple things:
1) Sometimes multiple jobs are associated with a single JobSet. I assume this is due to a failure and then restart but it is hard to verify this. Currently treating it as the job is not completed.
2) Sometimes no job is associated with the JobSet. I have only seen this once and am not sure if this was because the job was not started yet or automatically deleted after a completion. Currently treating it as the job is not completed.

RissyRan (approved):
LGTM! Thank you!

PR #578, opened 2025-01-24
[PyTorch] Remove unattended-upgrades

tengyifei:
Previously we stopped unattended-upgrades. But test runs indicated that unattended-upgrades still runs occasionally. Here we attempt to fix that by removing the binary altogether.

Tested: http://shortn/_Hr1JOwekCA


zpcore (approved):
LGTM! Curisous is `apt purge unattended-upgrades -y` a must in order to stop it?

tengyifei:
I don't know.

But this problem should really be fixed once and for all after this.

PR #580, opened 2025-01-25
refactor to use the e2e test script

michelle-yooh:
# Description

This PR refactors the maxtext_moe_gpu_e2e tests to use the MaxText e2e test script instead of the direct command. It requires https://github.com/AI-Hypercomputer/maxtext/pull/1191 to be merged first.

# Tests

Screenshot of the test result
http://shortn/_dNMK6pqBn9

bugs for the failed tests
pinned test on 2 A3 nodes: http://shortn/_WN4xgkG3Hj
pinned test on 2 A3+ nodes: http://shortn/_zKqVme3CFO

# Checklist

### DIFF
--- a/dags/sparsity_diffusion_devx/maxtext_moe_gpu_e2e.py
+++ b/dags/sparsity_diffusion_devx/maxtext_moe_gpu_e2e.py
@@ -23,14 +23,10 @@ from dags import composer_env
 from dags.common import test_owner
 from dags.common.vm_resource import XpkClusters, DockerImage
 from dags.multipod.configs import gke_config
-from xlml.utils.gpu import scale_up_a3_cluster, scale_down_a3_cluster
 
 
 # Run once a day at 11 am UTC (3 am PST)
 SCHEDULED_TIME = "0 11 * * *" if composer_env.is_prod_env() else None
-
-# Number of nodes on A3 cluster to be scaled up to
-A3_NUM_NODES = 3
 
 SCANNED_CHECKPOINT = "gs://ml-auto-solutions/output/sparsity_diffusion_devx/maxtext/chained_tests_mixtral-8x7b_nightly-2025-01-09-05-00-18//8x7b/scanned_ckpt/0/items"
 UNSCANNED_CKPT_PATH = "gs://ml-auto-solutions/output/sparsity_diffusion_devx/maxtext/chained_tests_mixtral-8x7b_nightly-2025-01-09-05-00-18//unscanned_ckpt/checkpoints/0/items"
@@ -55,15 +51,6 @@ def run_maxtext_tests():
   }
 
   for model, (test_script, nnodes) in test_models_gpu.items():
-    pinned_a3_gpu = gke_config.get_maxtext_end_to_end_gpu_gke_test_config(
-        time_out_in_min=90,
-        test_name=f"{test_name_prefix}-pinned-{model}",
-        run_model_cmds=(test_script,),
-        num_slices=nnodes,
-        cluster=XpkClusters.GPU_A3_CLUSTER,
-        docker_image=DockerImage.MAXTEXT_GPU_JAX_PINNED.value,
-        test_owner=test_owner.MICHELLE_Y,
-    ).run()
     pinned_a3plus_gpu = gke_config.get_maxtext_end_to_end_gpu_gke_test_config(
         time_out_in_min=90,
         test_name=f"{test_name_prefix}-pinned-{model}",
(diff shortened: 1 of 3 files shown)


dags/sparsity_diffusion_devx/maxtext_moe_gpu_e2e.py
```diff
@@ -16,61 +16,197 @@
...
+                )
+            ),
+        ],
+        cwd=tmpdir,
+    )
+    assert result.exit_code == 0, f"Command failed with code {result.exit_code}"
+
+
+def run_maxtext_tests(dag: models.DAG):
+  test_name_prefix = "maxtext"
```
parambole, line 130: @michelle-yooh -Is it possible to move this code to a utils file, as it is being used in multiple places[https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/dags/multipod/maxtext_gpu_end_to_end.py] and has the potential to be used in places where we need A3 tests? Additionally, in terms of the tests, do we need to have them for each A3 generation?

dags/sparsity_diffusion_devx/maxtext_moe_gpu_e2e.py
```diff
@@ -16,61 +16,197 @@
...
+  gcloud_command = (
+      f"gcloud container clusters resize {cluster_name}"
+      f"  --quiet --region {region}"
+      f"  --node-pool {node_pool}"
+      f"  --num-nodes {num_nodes}",
   )
+  return gcloud_command
+
+
+def wait_for_cluster_ready():
```
RissyRan, line 67: I feel those could be refactored a little bit and put as helper function in https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/xlml/utils/gpu.py? including `wait_for_cluster_ready`, `scale_up_cluster` and `scale_down_cluster`, `resize_a3_cluster`, etc?

michelle-yooh:
@RissyRan @parambole I've moved the functions to xlml/utils/gpu.py and verified that the resizing works as expected. http://shortn/_ttKViB2Hv7 (Please disregard the test failures as it used the latest MaxText image which doesn't have the test script yet)

xlml/utils/gpu.py
```diff
@@ -553,3 +557,76 @@ def wait_for_resource_deletion(operation_name: airflow.XComArg):
...
+        [
+            "bash",
+            "-c",
+            ";".join(
+                configure_project_and_cluster(
+                    XpkClusters.GPU_A3_CLUSTER.name,
+                    gke.zone_to_region(XpkClusters.GPU_A3_CLUSTER.zone),
+                )
+                + resize_a3_cluster(
+                    XpkClusters.GPU_A3_CLUSTER.name,
```
parambole, line 597: @michelle-yooh I believe there is an additional problem: running A3 tests affects the way clusters are set up. Given that this is the same approach used in the MaxText code, there could be a race condition where the cluster is being brought down while the tests are still running.

@yangyuwei correct me if I'm wrong, but the solution for this would be to bring up and down a new cluster or consolidate the tests into an existing DAG, which avoids rescaling a new cluster. Hence, I asked about the importance of running the tests on A3.

michelle-yooh: @parambole That's a valid concern. I had ignored it as the scheduled times for the DAGs are quite distant but I agree that there are still chances that other jobs are brought down in the mid way. 

@yangyuwei What were the reasons behind the resizing approach instead of keeping the fixed capacity for A3 cluster?

yangyuwei: +1 to the race condition @parambole brought up. We had a discussion in my original PR: https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/556.

Re "What were the reasons behind the resizing approach instead of keeping the fixed capacity for A3 cluster", it's also mentioned in my PR above, i.e., for A3 specifically, there is no way to keep the cluster running at a fixed number of nodes permanently at the moment, the cluster can be scaled down at any time without notice due to the fact that it's not being used. A better long-term solution is to create the cluster with autoscaling, but XPK doesn't support scheduling workloads on an autoscaling cluster at this point.

yangyuwei: I'd suggest considering two options proposed here: https://github.com/GoogleCloudPlatform/ml-auto-solutions/pull/556#discussion_r1917170528.

michelle-yooh: Merging it with maxtext_gpu_end_to_end defeats the purpose of having separate DAG for GPU MoE. And, it's also hard to come up with another sensible DAG only for A3. 

Now I'm with Param asking the importance of running the tests on A3. So, I'd like to suggest to drop A3 tests for this PR. We can implement the second option (or wait for XPK's support for autoscaling cluster) if we find the need for A3 tests but it seems out of scope of this PR anyway. 

@parambole @yangyuwei what do you think?

parambole: @michelle-yooh SGTM

yangyuwei: Sure. Sounds good!

yangyuwei (approved):

parambole (approved):

PR #596, opened 2025-02-04
Fix cluster name for a3ultra recipes

gunjanj007:
# Description

Fix cluster name in a3ultra

# Tests

# Checklist


dags/map_reproducibility/utils/common_utils.py
```diff
@@ -116,12 +116,12 @@ def helm_apply_cmds(
     recipe_repo_root,
     docker_image,
     aotc: bool = False,
+    cluster_name: str = "a3plus-benchmark",
```
cneel8986, line 119: Do we need to give a default cluster name which is internal

gunjanj007: We expose cluster names in all the DAGs, this is the only way to trigger DAGs in XLML right now

cneel8986 (approved):

PR #632, opened 2025-03-24
Reduce MaxText profiling tests

bvandermoon:
# Description

* Disable the MaxText/Vertex TB integration test
* Reduce the scope of the MaxText profiling test
  * Only check that the XPlane is saved to GCS instead of running assertions on the profile itself

The reason for this change is that the profiling work is not owned by us anymore. Note that we will likely come back and remove the Vertex DAG completely. But for now, we will just stop it from being scheduled.

# Tests

Will rely on the next Airflow run since these tests are already failing and aren't urgently needed

# Checklist

### DIFF
--- a/dags/multipod/maxtext_profiling.py
+++ b/dags/multipod/maxtext_profiling.py
@@ -48,12 +48,6 @@ with models.DAG(
         f" run_name=$RUN_NAME base_output_directory={base_output_directory}"
         f" dataset_path={dataset_path} profiler=xplane steps=20",
         f"gsutil cp -R {base_output_directory}/$RUN_NAME/tensorboard .",
-        "pip3 uninstall -y tbp-nightly",
-        "pip3 uninstall -y tensorboard_plugin_profile",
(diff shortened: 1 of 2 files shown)


gobbleturk:
Do we have unit tests or e2e tests that checks that profiler=xplane successfully works with maxtext (perhaps @hengtaoguo knows?)

bvandermoon:
> Do we have unit tests or e2e tests that checks that profiler=xplane successfully works with maxtext (perhaps @hengtaoguo knows?)

We do have [profiler unit tests](https://github.com/AI-Hypercomputer/maxtext/blob/main/MaxText/tests/profiler_test.py) in MaxText. They test that the profiler gets activated/deactivated, but they don't ensure that the XPlane actually gets written to GCS.

I'll keep the `maxtext_profiling.py` test active and just remove the extra profiling-specific assertions. That should help with our testing setup. That way we will only test the basic assertion of if the XPlane gets saved to GCS properly. Thanks Matt

gobbleturk (approved):

PR #638, opened 2025-03-26
[do not merge]turn on a3ultra benchmarking dag

Doris26:
# Description
remove the temporary code to use local file path since the new helm templated is fixed in this [CL](https://ai-hypercomputer-benchmarks-review.git.corp.google.com/c/reproducible-benchmark-recipes/+/3100). 

# Tests
https://screenshot.googleplex.com/3oGsamsFdbTVDaH
Run success and is reading recipes from internal-gpu-recipes
data shown correctly here: https://screenshot.googleplex.com/SoY9cxsBHgkqdZz

# Checklist

- [ X] I have performed a self-review of my code.
- [ X] I have necessary comments in my code, particularly in hard-to-understand areas.
- [ X] I have run one-shot tests and provided workload links above if applicable. 
- [ X] I have made or will make corresponding changes to the doc if needed.


richardsliu (approved):

dags/map_reproducibility/utils/common_utils.py
```diff
@@ -220,10 +220,9 @@ def helm_apply_cmds_internal_run(
   if aotc:
     set_aotc = " --set-string workload.aotc=true "
 
-  if hypercomputer == "a3mega":
-    helm_template_path = f"/home/airflow/gcs/dags/dags/map_reproducibility/helm-charts/{hypercomputer}/{framework}-training"
-  else:
-    helm_template_path = f"{recipe_repo_root}/src/helm-charts/{hypercomputer}/{framework}-training"
+  helm_template_path = (
+      f"{recipe_repo_root}/src/helm-charts/{hypercomputer}/{framework}-training"
+  )
```
polydier1, line 225: the change seems to be a better way to access the helm chart, but I am not sure is why we have specific path for a3mega orevuiyskt, have we tested all of the a3mega recipe already? (I would prefer Gunjan to take a look and confirm )

Doris26: The old code is to access my local helm charts. as there was a bug in the a3mega chart. 
[CL](https://ai-hypercomputer-benchmarks-review.git.corp.google.com/c/reproducible-benchmark-recipes/+/3100).

Steven fixed it so I need to change the code back :D


Current way is official way to pull the chart. Thank you!

stingram: Hey @Doris26 , just want to confirm, were you able to run all the a3mega DAGs and check they completed successfully? If yes, then we can approve. Thanks!

dags/map_reproducibility/internal_runs/a3mega_maxtext_benchmarking_dags.py
```diff
@@ -21,34 +21,33 @@
...
-    "recipes/a3mega/a3mega_llama3.1-8b_8gpus_bf16_maxtext.yaml": Schedule.DAILY_6PM_EXCEPT_THURSDAY,  # < 5mins
-    "recipes/a3mega/a3mega_llama3.1-8b_8gpus_fp8_maxtext.yaml": Schedule.DAILY_6PM_EXCEPT_THURSDAY,
-    "recipes/a3mega/a3mega_llama3.1-8b_16gpus_bf16_maxtext.yaml": Schedule.DAILY_6PM_EXCEPT_THURSDAY,
-    "recipes/a3mega/a3mega_llama3.1-8b_16gpus_fp8_maxtext.yaml": Schedule.DAILY_6PM_EXCEPT_THURSDAY,
-    # a3mega_mixtral-8x7 image issue
-    # "recipes/a3mega/a3mega_mixtral-8x7b_8gpus_bf16_maxtext.yaml": Schedule.DAILY_6PM_EXCEPT_THURSDAY,
-    # "recipes/a3mega/a3mega_mixtral-8x7b_8gpus_fp8_maxtext.yaml": Schedule.DAILY_6PM_EXCEPT_THURSDAY,
-    # "recipes/a3mega/a3mega_mixtral-8x7b_16gpus_bf16_maxtext.yaml": Schedule.DAILY_6PM_EXCEPT_THURSDAY,
-    # "recipes/a3mega/a3mega_mixtral-8x7b_16gpus_fp8_maxtext.yaml": Schedule.DAILY_6PM_EXCEPT_THURSDAY,
+    "recipes/a3mega/a3mega_llama3.1-8b_8gpus_bf16_maxtext.yaml": Schedule.DAILY_PST_6PM_EXCEPT_THURSDAY,  # < 5mins
```
wang2yn84, line 30: Just curious, is there any reason why we exclude Thursday?

wang2yn84 (approved):
Thank you!

dags/map_reproducibility/internal_runs/a3mega_maxtext_benchmarking_dags.py
gunjanj007: Thursdays are reserved for the external runs for recipes published

dags/map_reproducibility/utils/common_utils.py
```diff
@@ -434,6 +460,7 @@ def get_cluster(hardware: str = "a3ultra"):
     return "gke-a3ultra-bm-map-3", "europe-west1"
   if hardware == "a4":
     return "map-a4-gke", "us-central1"
+    return "gke-a3ultra-bm-map-3", "europe-west1"
```
gunjanj007, line 463: this is a4 cluster, why are you changing this??

dags/map_reproducibility/utils/common_utils.py
```diff
@@ -224,11 +247,13 @@ def helm_apply_cmds_internal_run(
   if aotc:
     set_aotc = " --set-string workload.aotc=true "
 
-  if hypercomputer == "a3mega":
+  if test_run:
```
gunjanj007, line 250: what is test_run doing here? why have you removed  if hypercomputer == "a3mega" herer?

gunjanj007 (changes requested):

dags/map_reproducibility/utils/common_utils.py
Doris26: This is because internal runs maintain a complete set of recipes and helm charts here, so if a non_test run occurs, it reads from there. https://source.corp.google.com/h/jax3p-gpu-benchmarking/internal-gpu-recipes

This is the function: helm_apply_cmds_internal_run.
The reason why we maintain completely different helm charts is because we encountered an issue where XLA_PYTHON_CLIENT_MEM_FRACTION = 0.98 works for the external recipe and image, but it does not work for the jax_stable_stack_nightly image which is used for the internal recipe runs. so we need to maintain a different set of recipes and helm charts : )

gunjanj007 (approved):

PR #672, opened 2025-04-14
Add option to provide custom XPK branch

abhinavclemson:
# Description

Add option to provide custom XPK branch. 
* Currently, XPK new version release interval is 2-3 weeks, it causes a long turn around period.

# Checklist

### DIFF
--- a/xlml/utils/xpk.py
+++ b/xlml/utils/xpk.py
@@ -26,6 +26,7 @@ from xlml.apis import metric_config
 from xlml.utils import gke
 from dags.common.vm_resource import GpuVersion
 
+MAIN_BRANCH = "main"
 # Duration = past 7 days
 LOGGING_URL_FORMAT = (
     "https://pantheon.corp.google.com/logs/query;"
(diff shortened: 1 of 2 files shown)


xlml/utils/xpk.py
```diff
@@ -41,12 +41,19 @@
 )
 
 
-def get_xpk_setup_cmd(tmpdir):
-  return [
+def get_xpk_setup_cmd(tmpdir, branch: str = ""):
+  if not branch:
+    branch = "v0.4.1"
```
RissyRan, line 46: I think this is actually for GPUs only. 

Could you help set default branch as master (i.e. default value as main/master instead of empty str. And if `accelerator_type` == GPU, then v0.4.1.

Something like:

1) if `accelerator_type` == GPU, then v0.4.1;
2) else use `branch` (either default main/master, or user defined value)

cc @yangyuwei @parambole

RissyRan (commented):
Thanks for adding this! Just one comment.

xlml/utils/xpk.py
abhinavclemson: 1). For workload creation, if accelerator type has value in GpuVersion Enum, branch would switch to "v0.4.1" , but if user has provided their branch, it would override it.
2). By default branch will be now master.

xlml/utils/xpk.py
```diff
@@ -263,7 +282,11 @@ def wait_for_workload_completion(
 
 @task(trigger_rule="all_done")
 def clean_up_workload(
-    workload_id: str, project_id: str, zone: str, cluster_name: str
+    workload_id: str,
+    project_id: str,
+    zone: str,
+    cluster_name: str,
+    xpk_branch: str = "",
```
RissyRan, line 289: Set a global variable as `MAIN_BRANCH="main"`, and then xpk_branch=MAIN_BRANCH.

```
def get_xpk_setup_cmd(tmpdir, branch: str = ""):
  clone_branch = (
      f"git clone --branch {branch} https://github.com/AI-Hypercomputer/xpk"
      f" {tmpdir}/xpk"
  )
  cmds = [
      "set -xue",
      clone_branch,
      "pip install ruamel.yaml docker",
  ]
  return cmds
```

HDYT?

abhinavclemson: 1). Added a global `MAIN_BRANCH="main"` in `xlml/utils/xpk.py`
2). Updated `get_xpk_setup_cmd` with suggested change.
3). 
Replaced `xpk_branch: str = ""` with `xpk_branch: str = MAIN_BRANCH` in `xlml/utils/xpk.py`
Replaced `xpk_branch: str = ""` with `xpk_branch: str = xpk.MAIN_BRANCH`  in `xlml/apis/tasks.py`

RissyRan (approved):
Thank you!

PR #706, opened 2025-05-02
Upload tensorboard metrics to BigQuery for maxdiffusion tpu e2e test

hyeygit:
# Description

Currently the maxdiffusion tests only verify correctness (e.g. training loss smaller than certain threshold), not performance (e.g. step time). This PR uploads tensorboard metrics (including step time) to BigQuery, so team can set up downstream pipelines to consume the perf metrics and add alerting.

Other changes in this PR to facilitate tensorboard->BigQuery upload:
* Use `run_with_name_gen_and_quarantine` to generate consistent run_name and tensorboard file location. The generated `run_name` is passed to the underlying maxdiffusion script via the environment variable `JOBSET_NAME`. See [here](https://github.com/AI-Hypercomputer/maxdiffusion/blob/7284ca0351aaba9bd7e24bf239b79b40bb86b34a/src/maxdiffusion/pyconfig.py#L119) for where maxdiffusion consumes this env var.
* Added some options to XLML's name generation utils so the generated names match those in maxdiffusion. For example, maxdiffusion hardcodes the format of the tensorboard file location as `os.path.join(base_output_directory, run_name, "tensorboard", "")`
* Maxdiffusion generates more than one Tensorboard event files per run, so I have to update the XLML metric util to grab the **largest** tensorboard event files since that will contain the most metrics. 

Also renamed `maxdiffusion_e2e` to `maxdiffusion_tpu_e2e` since a GPU version will be added in the near future.

# Tests

Tested with my own dev Cloud Composer env.
7vg3iBU9ZzpXRPV

# Checklist


dags/sparsity_diffusion_devx/maxdiffusion_tpu_e2e.py
```diff
@@ -48,11 +52,24 @@
   quarantine_task_group = TaskGroup(
       group_id="Quarantine", dag=dag, prefix_group_id=False
   )
-  current_datetime = config.get_current_datetime()
   for accelerator, slices in maxdiffusion_test_configs.items():
-    cores = accelerator.rsplit("-", maxsplit=1)[-1]
     cluster = config.clusters[accelerator]
     for slice_num in slices:
+      # The concrete run_name will be generated at runtime in `run_with_name_gen_and_quarantine`
```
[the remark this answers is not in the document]
hyeygit: Good idea. Done!

hyeygit:
cc @qinyiyan  - Hi Quinn, would you mind taking a look? Thank you!

qinyiyan (approved):
LGTM

qinyiyan:
> cc @qinyiyan - Hi Quinn, would you mind taking a look? Thank you!

The change looks good. I don't have the collaborator access to this repo.. Maybe Yixin or Richard can approve?

richardsliu (approved):

xlml/utils/name_format.py
```diff
@@ -33,22 +33,35 @@ def generate_run_name(benchmark_id: str) -> str:
 
 
 @task
-def generate_tb_file_location(run_name: str, base_output_directory: str) -> str:
+def generate_tb_file_location(
```
RissyRan, line 36: Probably we missed this before. It looks like we should enable a path `<base_output_directory>/whatever_folder/events.out.tfevents.*`. Otherwise, we could nested more and more different formats. No need to address in this PR, but just call out. cc @richardsliu

xlml/utils/metric.py
```diff
@@ -322,19 +326,29 @@ def get_gcs_file_location_with_regex(file_location: str) -> str:
...
-    return (
-        f"gs://{bucket_name}/"
-        f"{next(filter(file_path_regex.match, all_blobs_names))}"
-    )
-  except StopIteration:
+  if not matched_blobs:
     logging.warning(f"No objects matched supplied regex: {file_location}")
     return ""
 
+  if return_largest:
```
RissyRan, line 339: Trying to understand the logic of this largest file, i.e. should we process all files to get all metrics?

RissyRan (commented):
LGTM! Just have minor comments

xlml/utils/metric.py
hyeygit: Ideally yes, it'd be good to process all files, but probably as a future followup. Because:
1. For maxdiffusion-specific use case, the test always generates 2 tensorboard files, with the smaller one being a dummy (only 40 bytes). So picking the largest file works for our use case
2. metric.py currently is set up to expect only one tensorboard file, so some refactoring is required, which I think is better left as a separate PR.

xlml/utils/name_format.py
hyeygit: Ack, agree it'd be good to have a more generic version of this function.

xlml/utils/metric.py
RissyRan: > the smaller one being a dummy

Not sure why, but sg in this case we extract largest file.

RissyRan (approved):

PR #767, opened 2025-06-25
Add on_failure_listener plugin and always_fail.py example DAG for testing

severus-ho:
# Description

An Airflow Listener class to handle DAG run failures.
This listener specifically triggers actions when a DAG run fails.
It checks for the 'on_failure_alert' tag on the failed DAG. If the tag is present,
it proceeds to create a GitHub issue with details about the failed DAG run
and its failed tasks. The issue is assigned to the owners of the failed tasks

**Example issue:**
Example issues which made by triggering 'maxtext_end_to_end.py' in cienet-cmcs 
https://github.com/CIeNET-International/ml-auto-solutions-2/issues

# Tests

An new example DAG always_fail.py can easily test this plugin.

# Checklist

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -1,0 +1,48 @@
+The "on_failure_actions" plugin listens to the DAGs run results and takes action for failed DAG runs. Currently, the action is to file a GitHub issue for the failed DAG run.
+
+## Pre-requisites:
+To leverage the "on_failure_actions" plugin, ensure the following conditions are met:
+
+### 1.  **DAG Opt-In:** 
+Each DAG intended to utilize this feature **must include the `"on_failure_alert"` tag** within its DAG definition. DAGs without this specific tag will be ignored by the plugin's failure-handling logic, and no GitHub issue will be filed for their failures.
+
+    with DAG(
+        dag_id='my_critical_dag',
+        # ... other DAG parameters ...
+        tags=['data_ingestion', 'critical', 'on_failure_alert'], # <--- Add this tag
+    ) as dag:
+        # ... tasks ...
+
+### 2.  **GitHub Owner Mapping:** 
+For accurate issue assignment, ensure that the `owner` property defined for tests within your DAGs corresponds directly to valid **GitHub usernames**. The plugin will collect unique test owners from the failed DAG and attempt to assign the GitHub issue to these users.
+
+#### Example task definition
+    my_task = BashOperator(
+        task_id='process_data',
+        bash_command='...',
+        owner='github_username_here', # This should be a valid GitHub username
+    )
+
+    # Or
+
+    @task(owner='github_username_here') # This should be a valid GitHub username
+    def task_a():
+        pass
+
+## Configuration and Installation:
+1. From GCP console UI, Your Composer Env -> Tab -> Pypi packages -> Edit -> Add 'apache-airflow-providers-github' -> Save
+
+2. From GCP console UI, search for "Secret Manager", and add conn_id 'github_default' into Secret Manager.
+   1. key: airflow-connections-<composer_environment_name>-github_default
+   2. value:
+       {
+           "conn_type": "http",
+           "host": "https://api.github.com",
+           "password": "\<GitHub Personal Access Token\>"
+       }
+
+3. Composer -> Airflow configuration overrides -> Edit, reference: https://cloud.google.com/composer/docs/composer-1/configure-secret-manager
+   1. secrets | backend | airflow.providers.google.cloud.secrets.secret_manager.CloudSecretManagerBackend
+   2. secrets | backends_order | custom,environment_variable,metastore
+
+4. Upload 'on_failure_actions.py' to \<DAG Bucket\>/plugins/
--- a/plugins/on_failure_actions.py
+++ b/plugins/on_failure_actions.py
@@ -1,0 +1,195 @@
+import logging
+import os
+
+from airflow.exceptions import AirflowException
+from airflow.listeners import hookimpl
+from airflow.models import DagRun, TaskInstance
+from airflow.plugins_manager import AirflowPlugin
+from airflow.providers.github.hooks.github import GithubHook
+from github import Github
+
+from dags.common.vm_resource import Project
+from xlml.utils import composer
+from urllib import parse
+
+_PROJECT_ID = Project.CLOUD_ML_AUTO_SOLUTIONS
+_REPO_NAME = "GoogleCloudPlatform/ml-auto-solutions"
+
+
+def generate_dag_run_link(
+    proj_id: str,
+    dag_id: str,
+    dag_run_id: str,
+    task_id: str,
+):
+  airflow_link = composer.get_airflow_url(
+      proj_id,
+      os.environ.get("COMPOSER_LOCATION"),
+      os.environ.get("COMPOSER_ENVIRONMENT"),
+  )
+  airflow_dag_run_link = (
+      f"{airflow_link}/dags/{dag_id}/"
+      f"grid?dag_run_id={parse.quote(dag_run_id)}&task_id={task_id}&tab=logs"
+  )
+  return airflow_dag_run_link
+
+
+"""
+Airflow Listener class to handle DAG run failures.
+
+This listener specifically triggers actions when a DAG run fails.
+It checks for the 'on_failure_alert' tag on the failed DAG. If the tag is present,
+it proceeds to create a GitHub issue with details about the failed DAG run
+and its failed tasks. The issue is assigned to the owners of the failed tasks
+(excluding 'airflow' as an owner).
+
+TODO: Implement more sophisticated issue filing strategies beyond a single failure, such as:
+-   Consecutive Failures: Only file an issue if a DAG has failed for two or more
+    consecutive runs to reduce noise from transient issues.
+-   Reduced Pass Rate: File an issue if the current run failed AND the overall
+    pass rate of the past N (e.g., 10) runs for this DAG falls below a certain threshold
+    (e.g., 50%).
+"""
+
+
+class DagRunListener:
+
+  def __init__(self):
+    self.log_prefix = self.__class__.__name__
+
+  @hookimpl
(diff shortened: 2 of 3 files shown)


plugins/on_failure_actions.py
```diff
@@ -0,0 +1,123 @@
...
+It checks for the 'on_failure_alert' tag on the failed DAG. If the tag is present,
+it proceeds to create a GitHub issue with details about the failed DAG run
+and its failed tasks. The issue is assigned to the owners of the failed tasks
+(excluding 'airflow' as an owner).
+
+TODO: Implement more sophisticated issue filing strategies beyond a single failure, such as:
+-   Consecutive Failures: Only file an issue if a DAG has failed for two or more
+    consecutive runs to reduce noise from transient issues.
+-   Reduced Pass Rate: File an issue if the current run failed AND the overall
+    pass rate of the past N (e.g., 10) runs for this DAG falls below a certain threshold
```
shralex, line 49: Very nice ideas!

plugins/on_failure_actions.py
```diff
@@ -0,0 +1,123 @@
...
+    (e.g., 50%).
+-   Duplicate Issue Prevention: Before filing a new issue, check if a similar
+    issue for the same DAG (or error type) has already been filed within a recent
+    timeframe (e.g., the past 30 days) to avoid creating duplicate alerts.
+"""
```
shralex, line 54: In this case perhaps add a message to the existing issue (potentially rate-limit this) about the failure. 

Deduplication seems like a must -- otherwise we will now be flooded with github issues.

plugins/on_failure_actions.py
```diff
@@ -0,0 +1,123 @@
...
+        link = generate_dag_run_link(
+          proj_id=_PROJECT_ID,
+          dag_id=dag_run.dag_id,
+          dag_run_id=dag_run.run_id,
+          task_id=ti.task_id
+        )
+        body += f"Task {ti.task_id} failed, link:\n{link}\n"
+      assignees = list(
+        {
+          task.owner for task in dag_run.dag.tasks if task.owner != 'airflow'
```
shralex, line 102: If I understand correctly, this is owners of any task in this dag. Could it be better to take the owner of the task that failed, if defined, and otherwise the owner of the dag ?

shralex (commented):
Very elegant solution, thank you! And also thank you for adding the test.
I added a few comments. Most important seems the deduplication -- we could merge without this if we just want to test this in limited scope, but I think we must have deduplication to enable this for several dags.

plugins/on_failure_actions.py
severus-ho: We have weeded out all task owners whose tasks are successful. The remaining assignees are the owner of failed tasks.

plugins/on_failure_actions.py
severus-ho: We would search the 'open' and 'latest update_time' issue which has the same title, then append the new issue body at the bottom of the existing issue. If no issue found, we would create a new one.
In addition, the new set of assignees would replace the original set as updating the same issue.

plugins/on_failure_actions.py
```diff
@@ -0,0 +1,144 @@
...
+      if "on_failure_alert" not in dag_run.dag.tags:
+        logging.error(
+            f"[{self.log_prefix}] DAG {dag_run.dag_id} isn't "
+            f"'on_failure_alert' by tags. Return"
+        )
+        return
+      logging.error(
+          f"[{self.log_prefix}] DAG run {dag_run.dag_id} is 'on_failure_alert'"
+      )
+      title = f"[{self.log_prefix}] DAG {dag_run.dag_id} failed"
```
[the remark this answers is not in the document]
severus-ho: I pushed a new commit that issue title became "[DagRunListener] <DAG> <First failed task> failed".
About DAG owners, we need more internal discussion on this feature.

plugins/README.md
```diff
@@ -0,0 +1,48 @@
...
+        owner='github_username_here', # This should be a valid GitHub username
+    )
+
+    # Or
+
+    @task(owner='github_username_here') # This should be a valid GitHub username
+    def task_a():
+        pass
+
+## Configuration and Installation:
```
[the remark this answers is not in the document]
xibinliu: Hi Ran Ran, I only know the [ml-automation-solutions](https://pantheon.corp.google.com/composer/environments/detail/us-central1/ml-automation-solutions?referrer=search&inv=1&invt=Ab1z9w&project=cloud-ml-auto-solutions) prod env. Can you share what the other one is? If needed, we can deploy to both environments.

plugins/on_failure_actions.py
severus-ho: In addition, there is no owner attribute in DAG level. We could get only task owner from a DAG run

plugins/on_failure_actions.py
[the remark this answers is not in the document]
severus-ho: New update! Now the title format is "[DagRunListener] <DAG ID> <TEST NAME> failed"

plugins/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+    with DAG(
+        dag_id='my_critical_dag',
+        # ... other DAG parameters ...
+        tags=['data_ingestion', 'critical', 'on_failure_alert'], # <--- Add this tag
+    ) as dag:
+        # ... tasks ...
+
+### 2.  **GitHub Owner Mapping:** 
+For accurate issue assignment, ensure that the `owner` property defined for tasks within your DAGs corresponds directly to valid **GitHub usernames**. The plugin will collect unique task owners from the failed DAG and attempt to assign the GitHub issue to these users.
```
shralex, line 17: should this now be test instead of task here and below ?

plugins/on_failure_actions.py
```diff
@@ -0,0 +1,177 @@
...
+  return airflow_dag_run_link
+
+
+"""
+Airflow Listener class to handle DAG run failures.
+
+This listener specifically triggers actions when a DAG run fails.
+It checks for the 'on_failure_alert' tag on the failed DAG. If the tag is present,
+it proceeds to create a GitHub issue with details about the failed DAG run
+and its failed tasks. The issue is assigned to the owners of the failed tasks
```
shralex, line 43: also here - tests ?

plugins/on_failure_actions.py
```diff
@@ -0,0 +1,177 @@
...
+        link = generate_dag_run_link(
+            proj_id=str(_PROJECT_ID),
+            dag_id=dag_run.dag_id,
+            dag_run_id=dag_run.run_id,
+            task_id=ti.task_id,
+        )
+        if not first_failed_group_id:
+          first_failed_group_id = self.get_group_id(ti)
+        body += f"- **Failed Task**: [{ti.task_id}]({link})\n"
+        if ti.task.owner and ti.task.owner != "airflow":
```
shralex, line 139: are we taking the owner from the task or the task group ?

plugins/on_failure_actions.py
```diff
@@ -0,0 +1,177 @@
...
+        if (len(issues) > 0)
+        else None
+    )
+
+  def create_issue(self, client, title, body, assignees):
+    logging.error(f"[{self.log_prefix}] Create a new one")
+    repo = client.get_repo(full_name_or_id=_REPO_NAME)
+    repo.create_issue(title=f"{title}", body=f"{body}", assignees=assignees)
+
+  def update_issue(self, issue, body, assignees):
```
shralex, line 97: How would an updated issue look ? We should probably add a messages to the issue, rather than adding text to the existing message

shralex (approved):

plugins/on_failure_actions.py
shralex: its fine, but here's an alternative: https://github.com/AI-Hypercomputer/maxtext/issues/1912

plugins/on_failure_actions.py
severus-ho: https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/8b9e391b2eb600329d945704cacf8eecc24fabff/dags/multipod/configs/gke_config.py#L49-L81
https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/8b9e391b2eb600329d945704cacf8eecc24fabff/xlml/apis/task.py#L367-L387
Yes, the test_owner value would be passed into the owner of 'run_workload' task finally.

plugins/on_failure_actions.py
severus-ho: Now 'updating an issue' is to comment like the issue you provide.

plugins/on_failure_actions.py
```diff
@@ -0,0 +1,191 @@
...
+        return
+      logging.info(
+          f"[{self.log_prefix}] DAG run {dag_run.dag_id} is 'on_failure_alert'"
+      )
+      body = (
+          f"- **Run ID**: {dag_run.run_id}\n"
+          f"- **Execution Date**: {dag_run.execution_date}\n"
+      )
+      assignees = []
+      first_failed_group_id = ""
```
xibinliu, line 88: We should find all the failed task groups (first level groups only, which are also called "test cases"), not only the first one. So in theory, we can create issues for all the tests under the same DAG.

plugins/on_failure_actions.py
```diff
@@ -0,0 +1,191 @@
...
+            f"[{self.log_prefix}] Task {ti.dag_id}-{ti.task_id} failed"
+        )
+        link = generate_dag_run_link(
+            proj_id=str(_PROJECT_ID),
+            dag_id=dag_run.dag_id,
+            dag_run_id=dag_run.run_id,
+            task_id=ti.task_id,
+        )
+        if not first_failed_group_id:
+          first_failed_group_id = self.get_group_id(ti)
```
xibinliu, line 104: Are there cases where a task belongs to no groups and the group ID is none?

plugins/on_failure_actions.py
severus-ho: done

PR #821, opened 2025-08-17
Migrate 1B convergence tests to use v6e

gobbleturk:
# Description

Migrate 1B convergence tests from broken v4 cluster to v6e cluster. This additionally speeds up the tests, since trillium is 4x faster (each test only takes ~20 minutes, and we have 5 convergence tests, so ~100 cluster minutes total).

# Tests

Ran the 1B_convergence script manually

# Checklist

### DIFF
--- a/dags/multipod/maxtext_convergence.py
+++ b/dags/multipod/maxtext_convergence.py
@@ -49,8 +49,9 @@ with models.DAG(
 
   steps = 10200  # Half Chinchilla
   loss_threshold = 2.7
+  per_device_batch_size = 2.0 # Assuming 256 chips (v6e-256), we need global batch of 512
 
-  base_convergence_command = f"bash end_to_end/tpu/test_convergence_1b_params.sh OUTPUT_PATH={base_output_directory} DATASET_PATH={dataset_path} LOSS_THRESHOLD={loss_threshold} STEPS={steps}"
+  base_convergence_command = f"bash end_to_end/tpu/test_convergence_1b_params.sh OUTPUT_PATH={base_output_directory} DATASET_PATH={dataset_path} LOSS_THRESHOLD={loss_threshold} STEPS={steps} PER_DEVICE_BATCH_SIZE={per_device_batch_size}"
   convergence_tests = {
       "maxtext-convergence-bf16": ((base_convergence_command),),
       "maxtext-convergence-int8": (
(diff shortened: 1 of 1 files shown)


dags/multipod/maxtext_convergence.py
```diff
@@ -72,7 +73,7 @@
   maxtext_v4_config_tests = {}
   for test_name, run_command in convergence_tests.items():
     maxtext_v4_config_tests[test_name] = gke_config.get_gke_config(
```
shralex, line 75: can we also change variable names here and below to reflect v6e

shralex (approved):

dags/multipod/maxtext_convergence.py
shralex: I'm not sure how test names are constructed but they have v4 in them right now, can we change that:


maxtext-convergence-bf16-v4-128 |  
maxtext-convergence-int8-v4-128 |  
maxtext-convergence-subset-hosts-v4-128 |  
maxtext-convergence-grain-v4-128 |  
maxtext-convergence-hf-v4-128

gobbleturk: Good catch, renamed to v6e. I can't find where these names are generated either, I assume from the cluster name

dags/multipod/maxtext_convergence.py
RUEI4341, line 1: It looks like the **TPU_V6E_256_CLUSTER** name may have changed.
In [dags/common/vm_resource.py](https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/master/dags/common/vm_resource.py#L277), it's set to **bodaborg-v6e-256**, but in the GCP project, we only see [bodaborg-v6e-256-donotdelete-nl](https://pantheon.corp.google.com/kubernetes/clusters/details/us-central2/bodaborg-v6e-256-donotdelete-nl/details?invt=Ab534A&project=tpu-prod-env-large-adhoc)  
Are these referring to the same cluster?

gobbleturk: arg yes you are correct, I double checked this but still got it wrong...

It should be TPU_V6E_256_MLPERF_CLUSTER

https://github.com/GoogleCloudPlatform/ml-auto-solutions/blob/7165c1c37eacae1312d90ac8484f2c0f4ffd35d2/dags/common/vm_resource.py#L283

